PR #9, opened 2018-07-23
Add importer for Helsinki neighborhoods

igordavydsson:
Extends munigeo HelsinkiImporter to import only neighborhoods and
sub_districts as their children.


igordavydsson:
Closes in favor of #15 and changes made to `django-munigeo`

PR #14, opened 2018-09-17
Fix profile creation logic

igordavydsson:
Refs #2


juyrjola:
Hmm, let's think about how the profile API should reasonably work. For one user, there can be at most one profile object in existence, and the API will always return at most one profile object -- the one bound to the user authenticated by the bearer token.

For the user of the API, the distinction between `POST` and `PUT` becomes somewhat meaningless, because we won't allow adding more than one profile object anyway. In order to comply with RFC 2616, I think it'd be best if the profile data is manipulated only through `PUT` requests that point to the user's profile URI: `<prefix>/v1/profile/<uuid of user>/`. This implies that the `profile` resource is identified by the same UUID as the user.

Ping @tituomin, @tuomas777.

igordavydsson:
> Hmm, let's think about how the profile API should reasonably work. For one user, there can be at most one profile object in existence, and the API will always return at most one profile object -- the one bound to the user authenticated by the bearer token.

OK, that is the current implementation.

> For the user of the API, the distinction between `POST` and `PUT` becomes somewhat meaningless, because we won't allow adding more than one profile object anyway. In order to comply with RFC 2616, I think it'd be best if the profile data is manipulated only through `PUT` requests that point to the user's profile URI: `<prefix>/v1/profile/<uuid of user>/`. This implies that the `profile` resource is identified by the same UUID as the user.

OK, sounds reasonable, I'd say. Will implement that tomorrow then.

profiles/api.py
```diff
@@ -87,15 +93,15 @@ def get_queryset(self):
...
 
     def get_object(self):
-        try:
-            profile = self.get_queryset().get(user=self.request.user)
-        except Profile.DoesNotExist:
-            # TODO: create profiles when the user is created, not here.
-            profile = Profile.objects.create(user=self.request.user)
-        except (TypeError, ValueError):
-            raise Http404
+        return self.get_queryset().filter(user=self.request.user)
```
tuomas777: Does this actually work? Shouldn't `get_object()` return an object instance instead of a queryset?

igordavydsson: Hm, it works, but yeah, you would think that it needs to return an object...

tuomas777:
> Hmm, let's think about how the profile API should reasonably work. For one user, there can be at most one profile object in existence, and the API will always return at most one profile object -- the one bound to the user authenticated by the bearer token.
> 
> For the user of the API, the distinction between `POST` and `PUT` becomes somewhat meaningless, because we won't allow adding more than one profile object anyway. In order to comply with RFC 2616, I think it'd be best if the profile data is manipulated only through `PUT` requests that point to the user's profile URI: `<prefix>/v1/profile/<uuid of user>/`. This implies that the `profile` resource is identified by the same UUID as the user.
> 
> Ping @tituomin, @tuomas777.

Sounds good to me, of course the `uuid` in the url is a bit redundant, and one could consider having also something like `/v1/profile/me/` for accessing the current user's profile more easily.

I think supporting just PUT will make the implementation somewhat more messy and error-prone, as DRF don't have support for PUT object creation, but it is probably worth it considering usage from client POV. One thing to note also is that if the profile can be empty at start, and we really want to comply with RFC 2616, then the first PUT should return HTTP 201 instead of 200.

igordavydsson:
@juyrjola @tuomas777 PR updated.
To `POST`/`PUT` changes user has to send requests to `<prefix>/v1/profile/<uuid of user>/`.
If the user never had a profile and send the first requests to create it to `<prefix>/v1/profile/`, response status code is 201.

PR #17, opened 2018-09-21
Add basic tests for profiles

igordavydsson:
Basic tests + travis setup.

Refs #1

### DIFF
--- a/profiles/tests/test_profiles_api.py
+++ b/profiles/tests/test_profiles_api.py
@@ -1,0 +1,49 @@
+from rest_framework.reverse import reverse
+
+from profiles.models import Profile
+from profiles.tests.factories import ProfileFactory
+from profiles.tests.utils import get, post_create, put_update
+
+
+PROFILE_URL = reverse('profile-list')
+
+
+def get_user_profile_url(profile):
+    return reverse('profile-detail', kwargs={'user__uuid': profile.user.uuid})
+
+
+def test_unauthenticated_user_cannot_access(api_client):
+    get(api_client, PROFILE_URL, 401)
+
+
+def test_user_can_see_only_own_profile(user_api_client, profile):
+    other_user_profile = ProfileFactory()
+
+    data = get(user_api_client, PROFILE_URL)
+    results = data['results']
+    assert len(results) == 1
+    assert Profile.objects.count() > 1
+
+    get(user_api_client, get_user_profile_url(other_user_profile), status_code=404)
+
+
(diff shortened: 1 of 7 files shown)


igordavydsson:
or maybe @tituomin could roll Travis for us?

tituomin:
I activated Travis. Maybe push something to start the build?

igordavydsson:
@tituomin thanks! I will fix the problems :)

profiles/tests/test_profiles_api.py
```diff
@@ -0,0 +1,50 @@
...
+    assert Profile.objects.count() == 0
+
+    post_create(user_api_client, PROFILE_URL)
+
+    assert Profile.objects.count() == 1
+    profile = Profile.objects.latest('id')
+    assert profile.user == user_api_client.user
+
+
+def test_post_update_own_profile(user_api_client, profile):
```
tuomas777: maybe `test_put_update_own_profile` ?

profiles/tests/test_profiles_api.py
```diff
@@ -0,0 +1,50 @@
...
+    assert Profile.objects.count() == 1
+
+    user_profile_url = get_user_profile_url(profile)
+    phone_number_data = {'phone': '0461234567'}
+
+    put_update(user_api_client, user_profile_url, phone_number_data)
+
+    updated_profile = Profile.objects.get(user=profile.user)
```
tuomas777: You could use `profile.refresh_from_db()` which would also make sure, that we are dealing with the same profile instance.

profiles/tests/test_profiles_api.py
```diff
@@ -0,0 +1,50 @@
...
+    assert updated_profile.phone == phone_number_data['phone']
```
tuomas777: What happens to the other fields? If we want to comply with RFC 2616, they should be emptied I guess.

igordavydsson: so far other fiealds are empty 😬 
but good point, we'll check that it works that way in other cases. 👍

profiles/tests/test_profiles_api.py
igordavydsson: nice 👍

profiles/tests/test_profiles_api.py
igordavydsson: 👍

igordavydsson:
@tituomin @tuomas777 does it look ok to be merged?

Issue #13, opened 2018-09-06
Add nickname to profile


igordavydsson:
User should be able to add a nickname to his/her account after the profile has been registered.

PR #18, opened 2018-09-24
Add nickname to profile

igordavydsson:
Basic nickname that user can change at any point or keep blank.

Refs #13


profiles/tests/test_profiles_api.py
```diff
@@ -47,3 +47,23 @@ def test_put_update_own_profile(user_api_client, profile):
...
+        'divisions_of_interest',
+        'preferences'
+    )
+
+    user_profile_url = get_user_profile_url(profile)
+    profile_endpoint_data = get(user_api_client, user_profile_url)
+
+    assert len(profile_endpoint_data) == len(expected_fields)
+    for field_name in expected_fields:
+        assert field_name in profile_endpoint_data
```
tuomas777: Using `set`s would be more pro here, ie. `set(profile_endpoint_data.keys()) == set(expected_fields)` (or just `== expected_fields` if `expected_fields` was a `set` to begin with, which would be super pro).

igordavydsson: Fantastisch! 👍

igordavydsson: Feel so much more like a pro already! 😎

PR #19, opened 2018-09-24
Allow user to delete own profile

igordavydsson:
Refs #2

### DIFF
--- a/profiles/api.py
+++ b/profiles/api.py
@@ -3,7 +3,7 @@ import logging
 from django.utils.translation import ugettext_lazy as _
 from munigeo.models import AdministrativeDivision
 from parler_rest.serializers import TranslatableModelSerializer, TranslatedFieldsField
-from rest_framework import serializers, generics, viewsets, permissions
+from rest_framework import serializers, viewsets, permissions
 from rest_framework.exceptions import APIException
 from rest_framework.relations import RelatedField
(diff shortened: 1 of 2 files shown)


profiles/api.py
```diff
@@ -82,7 +82,7 @@ class Meta:
         model = Profile
 
 
-class ProfileViewSet(generics.RetrieveUpdateAPIView, viewsets.ModelViewSet):
+class ProfileViewSet(generics.RetrieveUpdateDestroyAPIView, viewsets.ModelViewSet):
```
tuomas777: Hmm, instead of `ApiView` and `ModelViewSet`, I think what you want here is a combination of `Mixins` and `GenericViewSet` like in [this](http://www.django-rest-framework.org/api-guide/viewsets/#example_3) example.

igordavydsson:
@tuomas777 thanks for the tip!
I think, if I use simple `ModelViewSet`, I will have pretty much the same result, considering how [it implements exactly that logic and all the needed methods](https://github.com/encode/django-rest-framework/blob/master/rest_framework/viewsets.py#L218)

tuomas777:
Yup, if you don't want to omit any of those, then `ModelViewSet` is definitely the way to go. I thought the point of the initial `RetrieveUpdateDestroyAPIView` was to omit some of the actions.

igordavydsson:
It was used to omit some actions, yes. But now we need all of those actions 😀

PR #34, opened 2019-07-26
Remove raven, add sentry-sdk

hkotkanen:
Removed `raven` and added `sentry-sdk`.

Haven't been able to test locally as I wasn't able to find the proper value for `SENTRY_DSN`...


open_city_profile/settings.py
```diff
@@ -37,10 +39,25 @@
 if os.path.exists(env_file):
     env.read_env(env_file)
 
+# Extract version (or commit hash) for sentry
 try:
-    version = raven.fetch_git_sha(checkout_dir())
+    version = subprocess.check_output(["git", "describe"]).strip()
```
igordavydsson: you could do `git describe --always`, it will fall back on the abbreviated commit hash, which is probably enough for sentry.

igordavydsson: would help you avoid nested try/except clauses 🙂

hkotkanen: Excellent point, much better way to do it. Will update.

PR #35, opened 2019-08-05
Add model change logging

hkotkanen:
REFS KUVA-183

This PR adds model instance change logging with `django-reversion`, which saves versions of serialized model instances in a separate table along with the user of the request. We use the middleware provided to automatically mark changes inside requests (excluding GET, HEAD, and OPTIONS) as belonging to a new revision.

`django-reversion` also provides admin integration for viewing, reverting to previous versions, and restoring deleted models.
image

New models must be marked for logging with `@reversion.register()` and bootstrapped with `createinitialrevisions` admin command.

NB! `django-reversion` works by detecting the signals fired by `save()`, so **bulk actions such as `Queryset.update()` won't be saved!**

### DIFF
--- a/profiles/api.py
+++ b/profiles/api.py
@@ -96,7 +96,7 @@ class ProfileViewSet(viewsets.ModelViewSet):
     renderer_classes = [JSONRenderer, BrowsableAPIRenderer]
 
     def get_queryset(self):
-        if self.request.user.is_staff:
+        if self.request.user.is_superuser:
             return self.queryset
         return self.queryset.filter(user=self.request.user)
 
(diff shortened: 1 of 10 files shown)


profiles/api.py
```diff
@@ -96,6 +96,8 @@ class ProfileViewSet(viewsets.ModelViewSet):
     renderer_classes = [JSONRenderer, BrowsableAPIRenderer]
 
     def get_queryset(self):
+        if self.request.user.is_staff:
```
igordavydsson: `.is_staff` might mean some very restricted permissions too, it's not `.is_superuser`. I would not probably allow any staff member to see all profiles by default.

hkotkanen: Yeah, very good point! Anyway we'll probably have to add more fine grained permission classes for youth officers and such.

igordavydsson: it should probably be something along the lines of `if self.request.user.has_permissions()`.

PR #36, opened 2019-08-07
Add relationship model to connect profiles

hkotkanen:
Refs KUVA-182

This PR adds a model for connecting two `Profile` instances (with `through` parameter of a `ManyToManyField`).

Model consists of a `representative` and a `representee`, and a field for recording the strength of authentication (in addition to suomi.fi *puolesta-asiointi*, a parent should be able to confirm their parenthood by either providing a signed proxy document or showing identification papers to a youth officer).

Profiles are also exposed in the admin interface with the representatives/representees shown as inlines.

Admin view for a parent `Profile` instance:
image

Admin view for a child `Profile` instance with two parents:
image

Still some TODOs missing:
- serialize relationships into response from profile endpoint?
- own endpoint for relationships?
- tests


open_city_profile/settings.py
```diff
@@ -174,6 +174,16 @@
     "DEFAULT_FILTER_BACKENDS": ("django_filters.rest_framework.DjangoFilterBackend",),
 }
 
+# User related settings
+
+REPRESENTATION_TYPE = (("custody", "Custodianship"),)
```
igordavydsson: these should be in `consts.py`?

profiles/models.py
```diff
@@ -29,6 +29,30 @@ def get_available_name(self, name, max_length=None):
...
+        max_length=30,
+        choices=settings.REPRESENTATION_TYPE,
+        default=settings.REPRESENTATION_TYPE[0][0],
+    )
+    confirmation_degree = models.CharField(
+        max_length=30,
+        choices=settings.REPRESENTATIVE_CONFIRMATION_DEGREE,
+        default=settings.REPRESENTATIVE_CONFIRMATION_DEGREE[0][0],
+    )
+    expiration = models.DateField(blank=True, null=True)
```
igordavydsson: does it need some default value or no?

open_city_profile/settings.py
hkotkanen: Moved to consts.

profiles/models.py
hkotkanen: I suspect that this information is something that comes from the suomi.fi puolesta-asiointi service when the relationship is confirmed by VRK data.

profiles/models.py
```diff
@@ -29,6 +31,28 @@ def get_available_name(self, name, max_length=None):
...
+    )
+    confirmation_degree = models.CharField(
+        max_length=30,
+        choices=REPRESENTATIVE_CONFIRMATION_DEGREE,
+        default=REPRESENTATIVE_CONFIRMATION_DEGREE[0][0],
+    )
+    expiration = models.DateField(blank=True, null=True)
+
+    def __str__(self):
+        return "{}".format(self.representative)
```
igordavydsson: maybe something like:
```python
return "{} - {}".format(self.representative, self.type)
```
?

profiles/models.py
igordavydsson: ok, we'll handle it later 👍

PR #37, opened 2019-08-07
Add model & admin for youth registration

hkotkanen:
Refs KUVA-181

This PR adds an app for youth customership with a `YouthProfile` model which is linked to a `Profile` instance.

TODO:
- ~~implement a `Permission` model that youths can use to request approval from their guardians.~~ this will be implemented in KUVA issue 189
- ~~tests~~

Admin view for `YouthProfile`:
image


open_city_profile/settings.py
```diff
@@ -213,3 +214,20 @@
                 "Please create a %s file with random characters to generate your secret key!"
                 % secret_file
             )
+
+# Youth profile settings
+
+GENDERS = (
```
igordavydsson: IMHO would be cleaner to keep these in `youths/consts.py`, if they are gonna be used only in this app, or in `open_city_profile/consts.py`, if they might be used elsewhere

youths/models.py
```diff
@@ -0,0 +1,61 @@
...
+class YouthProfile(models.Model):
+    # Required info
+    profile = models.OneToOneField(Profile, on_delete=models.CASCADE)
+    ssn = models.CharField(max_length=11)  # TODO ssn validation?
+    school_name = models.CharField(max_length=128)
+    school_class = models.CharField(max_length=10)
+    expiration = models.DateField(null=True, blank=True)
+
+    # Optional info
+    preferred_language = models.CharField(
```
igordavydsson: why not just use the [`language` field of the `Profile` model](https://github.com/City-of-Helsinki/open-city-profile/blob/master/profiles/models.py#L44)?

youths/models.py
```diff
@@ -0,0 +1,61 @@
...
+        max_length=32, choices=settings.LANGUAGES, default=settings.LANGUAGES[0][0]
+    )
+    volunteer_info = models.TextField(null=True, blank=True)
```
igordavydsson: please, avoid using `null=True` along with `blank=True` on `TextField` and other types of `CharField`. [this comment on SO](https://stackoverflow.com/a/52217122/7219196) gives the shortest explanation, why you should do it that way, other answer on that SO page give more detailed explanations. 🙂

igordavydsson: TLDR just use `blank=True` on you `CharField`s to avoid hiccups in the future

open_city_profile/settings.py
```diff
@@ -213,3 +214,20 @@
...
+    ("other", "Other"),
+    ("dont_know", "Don't know"),
+    ("rather_not_say", "Rather not say"),
+)
+
+ILLNESSES = (
+    ("diabetes", "Diabetes"),
+    ("epilepsy", "Epilepsy"),
+    ("heart_disease", "Serious heart or circulatory disease"),
+    ("serious_allergy", "Serious allergy"),
```
igordavydsson: is this list confirmed by someone from the city side? should it be `serious allergy` specifically? not any allergy?

youths/models.py
igordavydsson: maybe make the `YouthProfileAdmin` as inline of the `ProfileAdmin`? or can we have youth profiles without regular profiles?

utils/test_utils.py
```diff
@@ -13,6 +13,16 @@ def get(api_client, url, status_code=200):
     return response.data
 
 
+def patch_update(api_client, url, data=None, status_code=200):
```
igordavydsson: you created `utils` folder, because these test utils are now used in multiple apps? I guess, it is the question of personal preference, how to structure these things. I would probably put them in `open_city_profile/tests/utils` to signify that they are common for several apps. BUT a `utils` folder also works.

youths/api.py
```diff
@@ -0,0 +1,55 @@
...
+class YouthProfileSerializer(serializers.ModelSerializer):
+    class Meta:
+        exclude = ["id"]
+        depth = 1
+        model = YouthProfile
+
+    def create(self, validated_data):
+        # Membership always expires at the end of the season (30.6.).
+        # Signups before May expire in the summer of the same year, others next year.
+        if "expiration" in validated_data:
```
igordavydsson: should there also be some logic in the model's `save()` method, for the objects created through non-API methods? or maybe we could pass a function to the `default=` attribute of the `expiration` field on the model...

open_city_profile/settings.py
igordavydsson: settings usually have more to do with the internal of the project. now that I think about it, I should have put `CONTACT_METHODS` in `consts` too 😅

hkotkanen: excellent point. only in this app, so I'll make a `consts.py` file.

youths/models.py
hkotkanen: There were only fi, sv, and en in the `Profile` language choices, whereas this could have some rarer languages (some youth houses offer services in Somalian, Arabic, and maybe some other language as well). Not sure about this one

open_city_profile/settings.py
igordavydsson: resolved IRL

youths/models.py
hkotkanen: Will do. Many thanks for the heads up

open_city_profile/settings.py
hkotkanen: Unfortunately, there hasn't been anyone from the youth services available to confirm these, so I've made do with the documents prepared in their presence: [meeting notes](https://gitlab.com/City-of-Helsinki/dokumentaatio/ict-kehitys/wikis/meetingnotes/2019-04-08-nuorisoasiakkuus) and [UI flow wireframes](https://api.media.atlassian.com/file/0574012d-bb60-4301-8a80-253dee209b83/binary?client=02fbb8b3-7114-499e-bebc-cc4a1e0c1556&dl=true&token=eyJhbGciOiJIUzI1NiJ9.eyJpc3MiOiIwMmZiYjhiMy03MTE0LTQ5OWUtYmViYy1jYzRhMWUwYzE1NTYiLCJhY2Nlc3MiOnsidXJuOmZpbGVzdG9yZTpmaWxlOmE4YmQ0ZjljLTc4MmEtNGY2My05NWE2LTBhMDRlNzhlYmIxZSI6WyJyZWFkIl0sInVybjpmaWxlc3RvcmU6ZmlsZTo2OTc0YjZmNy04ZDdkLTRmMzYtOWE0Yy0zNjM2YTg5ZjAzZjYiOlsicmVhZCJdLCJ1cm46ZmlsZXN0b3JlOmZpbGU6MDU3NDAxMmQtYmI2MC00MzAxLThhODAtMjUzZGVlMjA5YjgzIjpbInJlYWQiXX0sImV4cCI6MTU2NTY4MTA4NCwibmJmIjoxNTY1NjgwMTI0fQ.4NNk6guC72qoIDBvNwsw8HWjwtpw4-9vuoVIppWJIv8).

hkotkanen: But yeah, these come from "spec" documents...

utils/test_utils.py
hkotkanen: You're right that these could go anywhere. I was perhaps prematurely anticipating that in the future we might add some common code for the apps to use and that they could also live in this `utils` folder. We can definitely move them to a directory that makes more sense to you, but if this is not entirely unsensible, I'd maybe leave it like this?

youths/api.py
hkotkanen: Good point, I guess these could also created with non-API methods... I'll move this logic to the `default` of the model field.

utils/test_utils.py
igordavydsson: yeah, as I said, matter of preference. common code could live in a separate directory under the name `utils`/`common` or it could live in the directory, where the settings file(s) are. I have seen both approaches in different projects, so totally up to you 👍

youths/models.py
igordavydsson: yeah, okay, let's keep this field. it probably should use a different tuple of options then

hkotkanen: About the inline admin widgets, good point. `YouthProfile` can't exist withouth `Profile`.

open_city_profile/settings.py
igordavydsson: probably should make labels translatable BTW

hkotkanen: I've now asked about this from the contact person I know, but they might take a while to answer. I suggest we leave it like t his for the time being and modify upon further comments.

igordavydsson: ok 👍

youths/models.py
hkotkanen: Profiles were inlined into `UserAdmin` and as far as I can tell, nested inlines aren't possible/easy without installing additional dependencies. I opted to expose `Profile`s as their own admin page, with `YouthProfile`s inlined.

utils/test_utils.py
hkotkanen: :+1: let's leave it like this

PR #39, opened 2019-08-15
Add django-ilmoitin notifications for LegalRelationship changes

hkotkanen:
Refs KUVA-199

Add `django-ilmoitin` and register a few notifications that are fired on changes to `LegalRelationship` model instances:

- If a relationship is created and it's not confirmed, notify the `representative` that their confirmation is needed.
- If a relationship is saved (created or updated), notify both `representative` and `representee` that an official relationship between them has been formed

Later we should probably define our own signals for more precise control (this PR uses the builtin `post_save` signal).


open_city_profile/settings.py
```diff
@@ -214,3 +216,7 @@
                 "Please create a %s file with random characters to generate your secret key!"
                 % secret_file
             )
+
+# Ilmoitin settings
+
+MAILER_EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
```
igordavydsson: should probably use env variables. [example from venepaikka](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/berth_reservations/settings.py#L35)

igordavydsson: also those mailer variables should probably be included as well. I think we will still use mailgun, [again venepaikka as example](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/berth_reservations/settings.py#L67)

profiles/notifications.py
```diff
@@ -0,0 +1,7 @@
+from django.utils.translation import ugettext_lazy as _
+from django_ilmoitin.registry import notifications
+
+notifications.register(
+    "relationship_confirmation_needed", _("Legal relationship confirmation needed")
```
igordavydsson: can we store the code strings as enum values? will be easier to change the Enum class and not change every place where the code string is used.

open_city_profile/settings.py
hkotkanen: Will do!

profiles/signals.py
```diff
@@ -0,0 +1,48 @@
...
+    if created and not confirmed:
+        try:
+            send_notification(
+                instance.representative.email,
+                "relationship_confirmation_needed",
+                instance.get_notification_context(),
+                instance.representative.language,
+            )
+        except (OSError, AnymailError) as e:
+            print(e)
```
igordavydsson: 🤔 weird that flake8 check did not catch this

hkotkanen: Yeah, damn flake8 leaving in these pesky print statements 😆

profiles/signals.py
```diff
@@ -0,0 +1,48 @@
...
+def relationship_saved_handler(sender, instance, created, **kwargs):
+    confirmed = instance.confirmation_degree != REPRESENTATIVE_CONFIRMATION_DEGREE[0][0]
+
+    # Only notify person whose confirmation is needed
+    if created and not confirmed:
+        try:
+            send_notification(
+                instance.representative.email,
+                "relationship_confirmation_needed",
+                instance.get_notification_context(),
```
igordavydsson: you need to add method `get_notification_context` to `LegalRelationship` model. [example from venepaikka](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/reservations/models.py#L202)

open_city_profile/settings.py
```diff
@@ -214,3 +222,17 @@
                 "Please create a %s file with random characters to generate your secret key!"
                 % secret_file
             )
+
+# Notification settings
```
igordavydsson: could you put these before `if "SECRET_KEY" not in locals():`, cause it's easier to keep that secret key generation logic at the end of the file, so you don't have to scroll over it every time you look through the settings

hkotkanen: Yes, definitely!

profiles/consts.py
```diff
@@ -1,4 +1,17 @@
 from django.utils.translation import ugettext_lazy as _
+from enumfields import Enum
+
+
+class NotificationType(Enum):
```
igordavydsson: well, consts are consts, enums are enums. `profiles/enums.py` is usually where those guys are.

hkotkanen: Indeed... splitting the file.

PR #40, opened 2019-08-22
Update django ilmoitin

hkotkanen:
Refs KUVA-199

Update to the newest `django-ilmoitin` version available - 0.1.3 - on PyPI.

Also add a `dummy_context` so that previewing the notification templates doesn't throw an exception.

### DIFF
--- a/profiles/notifications.py
+++ b/profiles/notifications.py
@@ -1,3 +1,4 @@
+from django_ilmoitin.dummy_context import dummy_context
 from django_ilmoitin.registry import notifications
 
 from .enums import NotificationType
@@ -10,3 +11,5 @@ notifications.register(
     NotificationType.RELATIONSHIP_CONFIRMED.value,
     NotificationType.RELATIONSHIP_CONFIRMED.label,
 )
+
+dummy_context.context.update({"relationship": None})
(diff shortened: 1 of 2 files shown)


profiles/notifications.py
```diff
@@ -10,3 +11,5 @@
     NotificationType.RELATIONSHIP_CONFIRMED.value,
     NotificationType.RELATIONSHIP_CONFIRMED.label,
 )
+
+dummy_context.context.update({"relationship": None})
```
igordavydsson: you can maybe make an in-memory `Relationship` object in the dummy context, so that you can get some of the fields in the preview.
```python
dummy_context.context.update({"relationship": Relationship(
    status=foo,
    other_field=bar
)})
```
at least that was the idea behind previews, not sure how it behaves in reality, as I haven't yet tested it 😅

hkotkanen: To create a dummy `Relationship`, I'd have to also create two dummy `Profiles` (which also requires a dummy `User`), and this didn't seem like a good place to do that.

igordavydsson: yeah, I was thinking, that it might be a problem. we can keept it the way it is now then 👍

PR #41, opened 2019-08-23
Remove ChoiceArrayField of illnesses, add permission fields to YouthProfile model

hkotkanen:
Add `BooleanField`s for the different illness choices and some text fields for extra information regarding ilnesses instead.

Other slight improvements:
- Group fields in Youth Profile admin
- Add Estonian and Russian to preferred languages choices
- Update season expiration date

This PR replaces the initial migration instead of adding a second migration, as the first migration referenced the custom ChoiceArrayField defined in `models.py`, which caused an exception when migrating because it's removed in this commit.

Firefox_Screenshot_2019-08-26T07-52-07 063Z

Refs None


youths/models.py
```diff
@@ -50,17 +32,27 @@ class YouthProfile(models.Model):
...
 
+    # Permissions
+    approved_by = models.ForeignKey(
+        Profile,
+        related_name="approved_youth_profiles",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+    )
+    approved_time = models.DateTimeField(null=True, blank=True)
```
igordavydsson: we probably want to disable the input for this field. you could set `editable=False`, and then mark the field as readonly in `admin.py`: `readonly_fields = ("approved_time",)`

hkotkanen: Good point. Updated.

PR #42, opened 2019-09-17
Graphql query and mutations for youth profiles

hkotkanen:
Refs OM-8, OM-9, OM-13, OM-14, OM-24

### YouthProfile model changes
- Approver is now only an email field to which a notification requesting confirmation is sent
- Store a token (`approval token`) which is used to validate whether the profile can be approved
- Store the timestamp when the latest notification was sent (could be used to calculate whether approval token has expired). This way, sent notifications are logged with `reversion`.

### GraphQL queries and mutations
Creation and update mutations of the youth profiles are implemented for logged in users. Non-admins can create a youth profile for themselves (and update their own youth profile), and admins can provide a `Profile` id to create and update a youth profile for anyone.

Update mutation accepts a `resendRequestNotification` field, which triggers a refresh of the approval token (invalidates the old one) and resends the notification.

With an "approval token" an specific youth profile's data can be queried ("This is what you are about to approve") and approved without signing in through tunnistamo.

### Temporary approval token
When creating a youth profile, a token is created and a notification is sent to the `approver_email` address. The message will/should include a link with the token included (most likely as a query param). The temporary token is a `uuid4`, and it can be used (without signing in) to query the data of the profile (providing the token as a variable to the query) and to mark it confirmed (by providing the token as an input to the mutation).

When a youth profile is mutated using the token, the stored token is invalidated by removing it.


youths/schema.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+PreferredLanguage = graphene.Enum(
+    "PreferredLanguage", [(l[0].upper(), l[0]) for l in LANGUAGES]
+)
+Gender = graphene.Enum("Gender", [(g[0].upper(), g[0]) for g in GENDERS])
+
+
+# Abstract base fields
+class YouthProfileFields(object):
```
igordavydsson: python 3 uses `object` as default base for all classes in nothing else is specified, so you can drop it from this line 🙂

igordavydsson: but maybe you can use `graphene.InputObjectType` as base here, then another input class will just extend this?

hkotkanen: I originally used `AbstractType` but it's deprecated so I followed the update [guide](https://github.com/graphql-python/graphene/blob/master/UPGRADE-v2.0.md#abstracttype-deprecated). You make an excellent point though, since the point of the AbstractType was that it can be mounted as an `InputObjectType` or a `ObjectType`, but since we use it only as an input, there's no need to inherit from the generic `object`.

youths/models.py
```diff
@@ -43,12 +43,10 @@ class YouthProfile(models.Model):
...
     # Permissions
-    approved_by = models.ForeignKey(
-        Profile,
-        related_name="approved_youth_profiles",
-        null=True,
-        blank=True,
-        on_delete=models.SET_NULL,
+    approver_email = models.EmailField()
+    approval_token = models.CharField(max_length=32, blank=True, editable=False)
+    approval_notification_sent = models.DateTimeField(
```
igordavydsson: `sent` sounds like ` boolean, maybe `approval_notification_timestamp`?

youths/notifications.py
```diff
@@ -0,0 +1,9 @@
+from django_ilmoitin.dummy_context import dummy_context
+from django_ilmoitin.registry import notifications
+
+notifications.register(
```
igordavydsson: could you do these as enums. [example](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/enums.py#L5)

youths/schema.py
```diff
@@ -0,0 +1,194 @@
...
+from .utils import create_approval_token, mail_encrypted_approval_link
+
+
+class YouthProfileType(DjangoObjectType):
+    class Meta:
+        model = YouthProfile
+        exclude = ["id", "approval_token"]
+
+
+PreferredLanguage = graphene.Enum(
```
igordavydsson: let's do them in a more human-friendly format, like we did [here](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/schema.py#L41). at least the languages. since labels for gender have apostrophes and spaces, we can use their codes.

youths/schema.py
```diff
@@ -0,0 +1,194 @@
...
+
+    youth_profile = graphene.Field(YouthProfileType)
+
+    @login_required
+    def mutate(self, info, **kwargs):
+        input_data = kwargs.pop("youth_profile_data")
+        if (
+            info.context.user.is_superuser
+        ):  # FIXME should be something like user.is_staff_or_more?
+            profile = Profile.objects.get(pk=kwargs.pop("profile_id"))
```
igordavydsson: `kwargs.pop("profile_id")` might resolve to `None` and the `profile` will be `None` too. later on fetching `YouthProfile` will fail, if no profile passed, I suppose. maybe raise an exception, if admin is trying to update someone's profile without passing its `id`?

igordavydsson: `from graphql import GraphQLError`

youths/schema.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            profile = Profile.objects.get(user=info.context.user)
+
+        youth_profile, created = YouthProfile.objects.get_or_create(
+            profile=profile, defaults=input_data
+        )
+        if not created:
+            raise Exception("Youth profile already exists for this profile!")
```
igordavydsson: would be best to raise `GraphQLErrors` here everywhere

youths/schema.py
```diff
@@ -0,0 +1,194 @@
...
+        send_notification(
+            email=youth_profile.profile.email,
+            notification_type="youth_registration_approved",
+            context={"youth_profile": youth_profile},
+        )
+        return ApproveYouthProfile(youth_profile=youth_profile)
+
+
+class Query(graphene.ObjectType):
+    youth_profile = graphene.Field(YouthProfileType, profile_id=graphene.Int())
```
igordavydsson: graphene has `ID` scalar, which can be and `Int` or `String`, so `profile_id=graphene.ID()` would be more appropriate

youths/schema.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # TODO should we hide some fields (e.g. gender) from the approver?
```
igordavydsson: good point, you could do another `ObjectType` for this. e.g. `YouthProfileForApproverType` or something along this logic, but with a more succinct name 😅

youths/notifications.py
hkotkanen: Sure, was supposed to but forgot. Thanks for 👁 it

youths/schema.py
hkotkanen: Yes, very good point. Will change them all of them.

youths/schema.py
```diff
@@ -0,0 +1,216 @@
...
+    youth_profile_by_approval_token = graphene.Field(
+        YouthProfileType, token=graphene.String()
+    )
+
+    @login_required
+    def resolve_youth_profile(self, info, **kwargs):
+        if (
+            info.context.user.is_superuser
+        ):  # FIXME should be something like user.is_staff_or_more?
+            return YouthProfile.objects.get(profile_id=kwargs.get("profile_id"))
```
igordavydsson: probably also needs to raise `GraphQLError` if no `Profile` exists

youths/schema.py
```diff
@@ -0,0 +1,216 @@
...
+    def mutate(self, info, **kwargs):
+        input_data = kwargs.pop("youth_profile_data")
+        profile_id = kwargs.pop("profile_id", None)
+
+        # TODO should be something like is_staff_or_more
+        if info.context.user.is_superuser:
+            try:
+                profile = Profile.objects.get(pk=profile_id)
+            except Profile.DoesNotExist:
+                raise GraphQLError("Invalid profile id")
```
igordavydsson: could you make all of these translatable 😅?

youths/migrations/0002_auto_20190912_1536.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.2.3 on 2019-09-12 12:36
```
igordavydsson: could also give this migration a more meaningful name?

igordavydsson: `change_approver_fields` or something in that direction

youths/schema.py
hkotkanen: Indeed. Missed it here, thanks!

youths/schema.py
```diff
@@ -14,10 +14,11 @@
 from .consts import GENDERS, LANGUAGES
 from .enums import NotificationType
 from .models import YouthProfile
-from .utils import create_approval_token, mail_approval_link
 
 
 class YouthProfileType(DjangoObjectType):
+    gender = graphene.Field(graphene.String, source="gender")
```
hkotkanen: This field is not required in the django model so it can be blank. This raises an error because `graphene.Enum` expects it to have a value, if I understood correctly. This is a workaround I found from the discussion of [this issue](https://github.com/graphql-python/graphene-django/issues/474).

igordavydsson: what if you declare `YouthProfileType` after `Gender = graphene.Enum("Gender", [(g[0].upper(), g[0]) for g in GENDERS])`, and then use it for this `gender` field:
```
gender = Gender()
```
would that help with the non-required issue? I don't know for sure though 😅

hkotkanen: Luckily easy to test. Still gives the same error: `Expected a value of type \"Gender\" but received: `. Return data includes `"gender": null,`...

hkotkanen: With the workaround, the data returned is `"gender": "",`, but includes no errors.

hkotkanen: Even though this workaround gets rid of the error, it will probably cause a bit of trouble for the frontend: the underlying field is actually an enum with a couple of choices (+ no value), but the data returned is just a string. While the frontend will still provide the value as an enum to the mutation, I don't think it's clear how it should show the returned data.

youths/tests/test_youth_profiles_graphql_api.py
```diff
@@ -0,0 +1,314 @@
...
+    query = """
+        {
+            youthProfile(profileId: 1) {
+                ssn
+                gender
+                schoolClass
+            }
+        }
+    """
+    expected_data = OrderedDict([("youthProfile", None)])
```
igordavydsson: you can avoid using OrderedDict and make tests slightly more readable by casting the result of `gql_client.execute()` to a `dict`:
```
expected_data = {"youthProfile": None}
executed = anon_user_gql_client.execute(query, context_value=request)
```

hkotkanen: Thank you! I'll change them

youths/schema.py
hkotkanen: I suggest we leave it like this for now and wait for feedback from the content specialists whether this is necessary.

youths/schema.py
hkotkanen: I suggest we leave it as is for now, record it as an issue and fix in a later PR.

youths/schema.py
igordavydsson: yeah, sure thing 👍

youths/schema.py
```diff
@@ -0,0 +1,218 @@
...
+    @login_required
+    def mutate(self, info, **kwargs):
+        input_data = kwargs.get("youth_profile_data")
+        profile_id = kwargs.get("profile_id")
+
+        if info.context.user.is_superuser:
+            try:
+                profile = Profile.objects.get(pk=profile_id)
+            except Profile.DoesNotExist:
+                raise GraphQLError(_("Invalid profile id."))
```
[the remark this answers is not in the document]
hkotkanen: I've been operating under the assumption that a superuser or a staff user can *not* have a Profile nor a youth profile. The profiles are associated with private citizens, whereas user objects with `is_staff` or `is_superuser` are strictly related to the official positions they hold.

youths/schema.py
```diff
@@ -0,0 +1,218 @@
...
+        youth_profile.save()
+
+        if resend_request_notification:
+            youth_profile.approval_token = uuid.uuid4()
+            send_notification(
+                email=youth_profile.approver_email,
+                notification_type=NotificationType.YOUTH_PROFILE_CONFIRMATION_NEEDED.value,
+                context={"youth_profile": youth_profile},
+            )
+            youth_profile.approval_notification_timestamp = datetime.now()
```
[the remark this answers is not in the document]
hkotkanen: Thanks! Will change.

youths/schema.py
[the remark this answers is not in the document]
igordavydsson: yeah, I had the same idea. staff accounts should not have profiles. maybe, we should think how to better enforce this.

youths/apps.py
```diff
@@ -1,5 +1,15 @@
...
 class YouthsConfig(AppConfig):
     name = "youths"
+
+    def __init__(self, *args, **kwargs):
+        super(YouthsConfig, self).__init__(*args, **kwargs)
+        import youths.notifications  # noqa isort:skip
+
+    # def ready(self):
+    #     if settings.NOTIFICATIONS_ENABLED:
+    #         import profiles.signals  # noqa isort:skip
```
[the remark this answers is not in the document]
hkotkanen: Whoops... Good 👁

youths/models.py
```diff
@@ -43,12 +44,12 @@ class YouthProfile(models.Model):
...
     # Permissions
-    approved_by = models.ForeignKey(
-        Profile,
-        related_name="approved_youth_profiles",
-        null=True,
-        blank=True,
-        on_delete=models.SET_NULL,
+    approver_email = models.EmailField()
+    approval_token = models.CharField(
+        max_length=36, blank=True, default=uuid.uuid4, editable=False
```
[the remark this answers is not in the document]
hkotkanen: I was considering it also, but ATM since there isn't an expiration time for tokens, they're invalidated by removing the token (setting the value to be `""`). `blank=True` and `unique=True` won't work together, right?

youths/schema.py
```diff
@@ -0,0 +1,218 @@
...
+                profile = Profile.objects.get(pk=profile_id)
+            except Profile.DoesNotExist:
+                raise GraphQLError(_("Invalid profile id."))
+        else:
+            try:
+                profile = Profile.objects.get(user=info.context.user)
+            except Profile.DoesNotExist:
+                raise GraphQLError(_("No profile found, please create one!"))
+
+        youth_profile = YouthProfile.objects.get(profile=profile)
```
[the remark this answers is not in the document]
hkotkanen: Good point, it should indeed.

youths/models.py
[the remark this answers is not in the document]
hkotkanen: If you guys recommend it, I could
- remove the `blank=True` and add `unique=True`
- add an expiration time
- always check an incoming token for expiration

I think this would mean that a token couldn't be removed so the stale token would be stored in the YouthProfile object until it's refreshed.

I recommend that we leave it as is and later pluck the token into it's own model.

youths/apps.py
```diff
@@ -1,5 +1,11 @@
 from django.apps import AppConfig
 
+# from django.conf import settings
```
igordavydsson: well, I guess also this should be deleted

hkotkanen: 🤦‍♂ yes

youths/models.py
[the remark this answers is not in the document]
tuomas777: > I was considering it also, but ATM since there isn't an expiration time for tokens, they're invalidated by removing the token (setting the value to be `""`). `blank=True` and `unique=True` won't work together, right?

Actually with the uniqueness constraint it needs to have also `null=True`. That would probably be my preferred solution, you could invalidate the token by setting it null.

youths/schema.py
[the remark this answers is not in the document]
tuomas777: ok, and in any case this is of course fine for now.

youths/models.py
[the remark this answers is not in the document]
hkotkanen: Your suggestion has been taken into consideration. The suggestion was rejected. Thank you for your suggestion.

😄 We'll refine the permissions/tokens feature in a future PR.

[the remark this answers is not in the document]
tuomas777: 😄

PR #44, opened 2019-09-30
Docker refactor

frwickst:
- Optimized Dockerfile for production
- Entrypoint for a complete project setup
- Single docker file project
- Settings from the environment file
- Creation of admin user
- Docker compose file with default network for linking multiple compose environments


utils/management/commands/add_admin_user.py
```diff
@@ -0,0 +1,24 @@
+from django.core.management.base import BaseCommand
+
+from users.models import User
```
igordavydsson: I think, it is more advisable to use `get_user_model` for referencing the user model:
```python
from django.contrib.auth import get_user_model

...

if not get_user_model().objects.filter(username=kwargs["username"]).count():
    get_user_model().objects.create_superuser(
        kwargs["username"], kwargs["email"], kwargs["password"]
    )
```

frwickst: Well, true that, it is better, I'll fix.

juyrjola:
Did you check the existing `createsuperuser` command?

frwickst:
You can also do it like this `echo "from django.contrib.auth import get_user_model; get_user_model().objects.create_superuser('admin', 'admin@example.com', 'pass')" | ./manage.py shell`
But yeah.. that is not the nicest, and hurts readability a lot.

frwickst:
Ah cool! Did not know, and _about time_! 👍

docker-entrypoint.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+until nc -z -v -w30 "$DATABASE_HOST" 5432
```
igordavydsson: is it the right port for the DB container? it is exposed on `5433` 🤔

frwickst: That is the port that it is exposed to the host with yes, but itself, inside the network, exposes on the 5432 port. So for other containers, `profile-db` will be accessible from `profile-db:5432` and from the host machine, it will be accessible from `localhost:5433`.

igordavydsson: ok, gotcha! 👍

pkalliok:
@Frodotus could you review this one, in order to get it merged?

PR #48, opened 2019-10-23
Refactor profile model

hkotkanen:
Earlier the `Profile` model didn't have a lot of useful fields, such as name (used to be stored in the `User` model) nor contact information (address, phone number etc.). This PR refactors the models so that both `BasicProfile` and `YouthProfile` inherit from a `ProfileBase` abstract model.

The abstract base model currently only includes a primary key UUID field, but my rationale was that when recording consents in the near future, we want a specific consent object to be able to be related to an arbitrary Profile object (ruling out Django's `ForeignKey` as that can only refer to a single type of model). This might or might not be the way to go.

We add some basic informational fields to the `BasicProfile`: first and last name, phone number, address (+zip and municipality), and date of birth. Because we store the dob in the basic profile now, the SSN field is removed from the `YouthProfile`.

**More commits are needed as this change completely breaks the APIs and thus most tests**: the GraphQL schemas (and REST API if it's seen as useful to keep) should be amended to reflect the changes introduced here.


voneiden:
Closing as stale

PR #49, opened 2019-10-25
OM-137 | Add first and last name to profile

pjlampil:
- Added fields `first_name` & `last_name` to `Profile`
- Added data migration that populates fields for existing objects from `User` if relationship exists
- Overridden `Profile`'s `save` method to populate fields from `User` if relationship exists
- Altered the GraphQL endpoints accordingly
- Added tests for model and altered GraphQL tests


profiles/models.py
```diff
@@ -59,6 +59,8 @@ def get_notification_context(self):
 @reversion.register()
 class Profile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE)
+    first_name = models.CharField(max_length=255, null=True, blank=True)
```
igordavydsson: It is not advisable to allow both `null=True` and `blank=True` on `CharField` or its descendants, as it can cause confusions and incosistencies with some empty values stored as blanks and some as nulls. Unfortunately, this was missed in earlier stages of the profile's development and now we have some `CharField` where both are allowed 😅 , like in the case of `username` a couple of rows down. Would be super neat, if somebody could fix those at some point. 🙏

profiles/migrations/0008_add_first_name_and_last_name_to_profile.py
```diff
@@ -0,0 +1,31 @@
...
+            model_name="profile",
+            name="first_name",
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+        migrations.AddField(
+            model_name="profile",
+            name="last_name",
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+        migrations.RunPython(populate_name_from_user_for_existing_profiles),
```
igordavydsson: I think you should explicitly specify `migrations.RunPython.noop` as the reverse operation here, or django won't allow running migrations backwards. or do I misremember? 🤔

profiles/migrations/0008_add_first_name_and_last_name_to_profile.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.3 on 2019-10-24 11:26
+
+from django.db import migrations, models
+
+
+def populate_name_from_user_for_existing_profiles(apps, schema_editor):
+    Profile = apps.get_model("profiles", "Profile")
+    for profile in Profile.objects.all():
+        if profile.user is not None:
```
igordavydsson: `if profile.user` would do the trick?

profiles/migrations/0008_add_first_name_and_last_name_to_profile.py
```diff
@@ -0,0 +1,31 @@
...
+            profile.first_name = profile.user.first_name or profile.first_name
+            profile.last_name = profile.user.last_name or profile.last_name
+        profile.save()
```
igordavydsson: probably want to save only those profiles that were actually modified through the `if` statement?

profiles/models.py
```diff
@@ -84,10 +86,14 @@ class Profile(models.Model):
         "self", through=LegalRelationship, symmetrical=False
     )
 
+    def save(self, *args, **kwargs):
+        if self.pk is None and self.user is not None:
+            self.first_name = self.user.first_name or self.first_name
```
igordavydsson: this would override any values passed by the user through an API, right? what was the use case for this?

igordavydsson: ah, probably for the initial User object creation. hm 🤔 I guess, we have to do this in the parts where we do `Profile.objects.create()`, cause having this in `save()` can cause unwatned overrides.

igordavydsson: or is it exactly the spec, that no matter what the user says, we check what we received from the IDP via Tunnistamo 😅

profiles/models.py
pjlampil: Yep. I was kind of following the convention as I'm not even sure if the case where those fields are null or blank is really valid. But these should be fixed (at some point).

profiles/models.py
pjlampil: The use case was to use users name as an initial value for profile's `first_name` and `last_name` fields. If statement is only going to truthy for new objects with related user present.

igordavydsson: oh, right, it checks for `pk`. 👍 more succinct would maybe be `if not self.pk and self.user:`?

profiles/models.py
igordavydsson: Could you drop `null=True` from these? at least new fields would be already safe 🙂 we can fix the old fields later

profiles/migrations/0008_add_first_name_and_last_name_to_profile.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [("profiles", "0007_add_legal_relationship_model")]
+
+    operations = [
+        migrations.AddField(
+            model_name="profile",
+            name="first_name",
+            field=models.CharField(blank=True, max_length=255, null=True),
```
igordavydsson: also from here 😉

PR #51, opened 2019-10-30
OM-141 | Profile query for berth

pjlampil:
- Fixed `Profile` not defaulting to `User` name if already set.
- Added query for fetching berth customer data with filtering, ordering and pagination.

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -87,7 +87,7 @@ class Profile(models.Model):
     )
 
     def save(self, *args, **kwargs):
-        if not self.pk and self.user:
+        if not self.pk and not (self.first_name and self.last_name) and self.user:
             self.first_name = self.user.first_name or self.first_name
             self.last_name = self.user.last_name or self.last_name
         super().save(*args, **kwargs)
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -3,11 +3,17 @@ from django.conf import settings
 from django.db.models import CharField, Value
 from django.db.models.functions import Concat
 from django.utils.translation import override
+from django.utils.translation import ugettext_lazy as _
+from django_filters import CharFilter, FilterSet, OrderingFilter
+from graphene import relay
+from graphene_django.filter import DjangoFilterConnectionField
 from graphene_django.types import DjangoObjectType
+from graphql import GraphQLError
 from graphql_jwt.decorators import login_required
 from munigeo.models import AdministrativeDivision
 from thesaurus.models import Concept
 
+from services.consts import SERVICE_TYPES
 from services.models import Service
 from services.schema import ServiceType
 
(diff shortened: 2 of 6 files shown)


profiles/models.py
```diff
@@ -87,7 +87,7 @@ class Profile(models.Model):
     )
 
     def save(self, *args, **kwargs):
-        if not self.pk and self.user:
+        if not (self.first_name and self.last_name) and not self.pk and self.user:
```
igordavydsson: this was in a nice own commit, now it is lumped with the changes it has nothing to do with 😅

igordavydsson: I would maybe start with checking, if the object has `pk` first, as it would probably already stop the logic for so many cases. then you could do a slightly "heavier" `not (self.first_name and self.last_name)`, and then check, if the object has the user set.

igordavydsson:
```python
if not self.pk and not (self.first_name and self.last_name) and self.user:
    ...
```

profiles/schema.py
```diff
@@ -50,10 +56,60 @@ def resolve_type(self, info, **kwargs):
...
+
+    count = graphene.Int()
+    total_count = graphene.Int()
+
+    def resolve_count(self, info):
+        return self.length
+
+    def resolve_total_count(self, info):
+        return self.iterable.model.objects.filter(
+            service__service_type=SERVICE_TYPES[1][0]
```
igordavydsson: I would put this in a util function, so that if one day those indices change, we want have to look for them everywhere. maybe `services/utils.py`: 
```python
get_berth_service():
    return SERVICE_TYPES[1][0]
```

you can then just call this function whenever you need.

profiles/schema.py
```diff
@@ -50,10 +56,60 @@ def resolve_type(self, info, **kwargs):
...
+    class Meta:
+        model = Profile
+        fields = ("first_name", "last_name", "nickname", "email", "phone", "language")
+
+    first_name = CharFilter(lookup_expr="icontains")
+    last_name = CharFilter(lookup_expr="icontains")
+    nickname = CharFilter(lookup_expr="icontains")
+    email = CharFilter(lookup_expr="icontains")
+    phone = CharFilter(lookup_expr="icontains")
+    language = CharFilter(lookup_expr="icontains")
```
igordavydsson: maybe it should be `exact`? I am not sure, but maybe. 🤔 it is default, so if you agree, you can just do `CharFilter()` here on this line.

profiles/tests/test_profiles_graphql_api.py
```diff
@@ -0,0 +1,181 @@
...
+            }
+        }
+    """
+
+    expected_data = {"berthProfiles": {"count": 1, "totalCount": 2}}
+
+    executed = superuser_gql_client.execute(
+        query, variables={"firstName": profile_2.first_name}, context=request
+    )
+    print(executed)
```
igordavydsson: forgot `print` statement 🙂

profiles/tests/test_profiles_graphql_api.py
```diff
@@ -0,0 +1,181 @@
...
+        }
+    """
+
+    expected_data = {
+        "berthProfiles": {
+            "edges": [{"node": {"firstName": "Bryan"}}, {"node": {"firstName": "Adam"}}]
+        }
+    }
+    executed = superuser_gql_client.execute(query, context=request)
+    print(executed)
```
igordavydsson: and here

profiles/models.py
pjlampil: > this was in a nice own commit, now it is lumped with the changes it has nothing to do with 😅

This was a mistake by me when squashing the commits. My bad. I noticed it when I was writing the tests for the actual task here.

profiles/schema.py
pjlampil: There is actually a task to use real enums with these (OM-155). I think that should fix it.

igordavydsson: ok, great!

profiles/schema.py
```diff
@@ -50,10 +56,60 @@ def resolve_type(self, info, **kwargs):
...
+            service__service_type=SERVICE_TYPES[1][0]
+        ).count()
+
+
+class ProfileFilter(FilterSet):
+    class Meta:
+        model = Profile
+        fields = ("first_name", "last_name", "nickname", "email", "phone", "language")
+
+    first_name = CharFilter()
```
igordavydsson: I actually meant `exact` makes sense for the `language` field, where you have only codes and it is probably best to get the code that matches excatly. other fields might be quite useful to be `icontains`.

PR #52, opened 2019-10-31
OM-147 | Modify service type and profile unique together

pjlampil:
- Add a constraint to `Service` to make it unique by `profile` and `service_type`
- Add a migration to remove possibly existing duplicates
- Modify resolver to throw error instead of failing silently in case of duplicate
- Modify resolver to return GraphQL error


services/tests/test_services_graphql_api.py
```diff
@@ -51,3 +53,33 @@ def test_normal_user_can_add_service_mutation(rf, user_gql_client):
...
+                    type
+                }
+            }
+        }
+        """
+    )
+    creation_data = {"serviceType": "BERTH"}
+    query = t.substitute(**creation_data)
+    expected_data = {
+        "addService": OrderedDict({"service": OrderedDict({"type": "BERTH"})})
```
igordavydsson: does it have to be `OrderedDict` here? you cast it to dict anyway a couple of lines later: `dict(executed["data"])`

services/models.py
```diff
@@ -10,6 +10,9 @@ class Service(models.Model):
     service_type = models.CharField(max_length=32, choices=SERVICE_TYPES, blank=False)
     created_at = models.DateTimeField(auto_now_add=True)
 
+    class Meta:
+        unique_together = ["profile", "service_type"]
```
igordavydsson: we usually use tuples for most listings in places like this. I think the reason is that it gives you that 0.0001 % increase in security against some weird runtime bugs 😅

pjlampil: Kinda makes sense, although official docs prefer list. Let's go with tuples.

PR #53, opened 2019-10-31
Add possibility to add multiple emails, phonenumbers and addresses

Frodotus:
Remove rest api
Add possibility to add multiple emails, phone numbers and addresses

Refs: OM-134


profiles/models.py
```diff
@@ -102,3 +106,29 @@ class DivisionOfInterest(models.Model):
...
+class Email(Contact):
+    email = models.EmailField(null=True, blank=True)
+    email_type = models.CharField(max_length=32, choices=EMAIL_TYPES, blank=False)
+
+
+class Address(Contact):
+    address = models.CharField(max_length=128, blank=False)
+    postal_code = models.CharField(max_length=5, blank=False)
+    city = models.CharField(max_length=64, blank=False)
+    country = models.CharField(max_length=64, blank=True)
```
klempine: Just occurred to me that we could store the country as a 2 character iso code. Perhaps even with this: https://pypi.org/project/django-countries/

PR #55, opened 2019-11-06
Docker update

igordavydsson:
It is a good practice to switch to non-root users by default,
and our images come with `appuser` already set up.


pjlampil:
Makes sense to me 👍

PR #56, opened 2019-11-06
OM-159 | Add sensitive data to profile

pjlampil:
Add `SensitiveData` model
- 1-to-1 relationship to profile
- encrypted field `ssn`


open_city_profile/settings.py
```diff
@@ -253,3 +254,7 @@
                 "Please create a %s file with random characters to generate your secret key!"
                 % secret_file
             )
+
+FIELD_ENCRYPTION_KEYS = [
+    "f164ec6bd6fbc4aef5647abc15199da0f9badcc1d2127bde2087ae0d794a9a0b"
```
igordavydsson: this should probably be set through an env variable?

klempine:
> 
> 
> Looks good! the only thing that bothers me is that we can easily forget to change the key in production. I have couple of ideas:
> 
>     1. force generation of new key when debug is false (i.e. check if the example key is in `FIELD_ENCRYPTION_KEYS`, when `DEBUG == False`)
> 
>     2. force generation of new key always, i.e. drop this key from the `docker-compose.env.yaml.example` and put there `FIELD_ENCRYPTION_KEYS=<GENERATE YOU OWN KEY>` or something along these lines.
> 
> 
> and there should be some instructions then in the `README` on how to generate those keys.

I created a separate ticket for this: https://helsinkisolutionoffice.atlassian.net/browse/OM-191. We'll handle the key generation etc there, since closing this PR doesn't require it.

PR #57, opened 2019-11-06
OM-169 | Switch profile model's pk from int ID to UUID

igordavydsson:
Several foreign key relations to the profile model make this
slightly more complex, than may be desired. However, migrations
handle those cases gracefully, migrating all the pointers both
forwards and backwards. The only data lost during this switch
is concepts and divisions of interest of the profile objects.

In the future, migrations can be squashed and all the data
migrations erased, making the migration files much lighter.


pjlampil:
And we should definitely squash migrations at some point.

PR #58, opened 2019-11-06
OM-167 | Fix Django admin view for YouthProfile

igordavydsson:
The fields have been heavily refactored, but the admin file
was missed. This commit makes it work again.


pjlampil:
True story, thanks!

PR #62, opened 2019-11-12
Configure GitLab CI pipeline

antti-mikael:


Dockerfile
```diff
@@ -7,23 +7,30 @@ ENV PYTHONUNBUFFERED 1
...
-    && rm -rf /var/cache/apt/archives
-
-COPY --chown=appuser:appuser requirements.txt /app/requirements.txt
-RUN pip install -U pip \
-    && pip install --no-cache-dir -r /app/requirements.txt
+        python3-gdal \
+    && pip install --no-cache-dir \
+        -r /app/requirements.txt \
+        -r /app/requirements-prod.txt \
+    && apt-cleanup.sh \
```
igordavydsson: @tuomas777 just had issues with running `pip install ...` later in the `Dockerfile` after this `apt-cleanup.sh` has been run. he only managed to make it work by moving `apt-cleanup.sh` to `development` and `production` sections

antti-mikael: That was fixed by adding `apt-install.sh build-essential` in the `development` stage. See the final revision of that PR: https://github.com/City-of-Helsinki/kukkuu/pull/8/commits/402a09cbe60a76e0ce8b84eab764954e3b57888e

igordavydsson: yes, we Talked with Tuomas about this. 👍

Dockerfile
```diff
@@ -7,22 +7,32 @@ ENV PYTHONUNBUFFERED 1
 WORKDIR /app
```
ruriky: Not related to your changes directly but when using helsinkitest images, WORKDIR /app is already set for you.

antti-mikael: I think it's better to be explicit here.

antti-mikael:
Formatting fixed and conflicts resolved.

PR #65, opened 2019-11-13
OM-80 | Permissions

pjlampil:
- Add `ServiceConnection` between `Profile` and `Service`
- Use Relay interface for service related GraphQL object types
- Add admin for `Service`
- Add `guardian` for object permissions
- Add `staff_required` decorator
- Change `berthProfiles` query to generic `profiles`
- Change `profile` query permissions
- Add `myProfile` query


services/models.py
```diff
@@ -22,8 +22,8 @@ def __str__(self):
 
 
 class ServiceConnection(models.Model):
-    profile = models.ForeignKey(Profile, on_delete=models.PROTECT)
-    service = models.ForeignKey(Service, on_delete=models.PROTECT)
+    profile = models.ForeignKey(Profile, on_delete=models.CASCADE)
+    service = models.ForeignKey(Service, on_delete=models.CASCADE)
```
igordavydsson: would it make sense to `PROTECT` here, to not allow deleting a service object if it has connected profiles? something @klempine might weigh in on

igordavydsson: profile fk would be `CASCADE`, yes.

profiles/schema.py
```diff
@@ -205,13 +206,34 @@ def mutate(self, info, **kwargs):
 
 
 class Query(graphene.ObjectType):
-    profile = graphene.Field(ProfileType)
+    profile = graphene.Field(
+        ProfileType,
+        id=graphene.Argument(graphene.UUID, required=True),
```
igordavydsson: our `ProfileType` is actually a node, and it will be exposed with graphene's [`GlobalID`](https://github.com/graphql-python/graphene/blob/master/graphene/relay/node.py#L28) as `id`. I think it will be simpler for the frontend to use that global ID for this `profile` query too. you can then use [this class method](https://github.com/graphql-python/graphene/blob/master/graphene/relay/node.py#L111) to resolve it. yeah, it's gonna be more verbose, but probably more consistent.

services/models.py
pjlampil: Yeah. `PROTECT` on `Service`, `CASCADE` on `Profile` but there should be something in the future to protect profiles connected to certain services (at least in certain state).

profiles/schema.py
pjlampil: Of course. I was blinded by ID's. Does it make any sense now?

PR #66, opened 2019-11-13
VEN-252 | Federated profiles

igordavydsson:
Berth-reservations' profiles will have extra fields and will act
as extension to these profiles. Apollo-gateway will merge the
schemas into one, so that the profile object at the gateway
level will have fields from both backends at the same time.
All of the extensions will be handled in the berth-reservations
repo, but open-city-profile needs these adjustments to make
it all work.

As a 🍒 on top: refactored the tests.


profiles/schema.py
```diff
@@ -120,7 +121,8 @@ class Meta:
         fields = ("address_type", "primary", "email")
 
 
-class ProfileType(DjangoObjectType):
+@key(fields="id")
+class ProfileNode(DjangoObjectType):
```
pjlampil: This naming is more like a convention?

igordavydsson: yeah, Tuomas and me been bumping into this when googling around. Usually the class or entity gets named after its interface, which are usually nodes. only if it does not `implement` anything, then it is a `Type`. so, Tuomas started followng this in kukkuu and I thin k we should try to follow it everywhere, and I will try to update the GraphQL guidelines.

PR #67, opened 2019-11-14
OM-198 / Added a field for birth date

klempine:
OM-198 / Added a field for birth date, changed the unit tests and added some GraphiQL documentation


youths/models.py
```diff
@@ -22,6 +22,7 @@ class YouthProfile(models.Model):
     profile = models.OneToOneField(
         Profile, related_name="youth_profile", on_delete=models.CASCADE
     )
+    birth_date = models.DateField()
```
igordavydsson: is it obligatory?

klempine: Yes, at least in my opinion. Without the birth date we cannot know how old the youth is. We have to know the age in order to validate certain things and contact the guardian for confirmation, if necessary.

PR #70, opened 2019-11-20
OM-199 | Add membership number for youth profile

klempine:
The membership number of a youth profile is simply the youth
profile model's PK padded with zeroes. I added a new property to
the model so that the property returns the padded number. The
length of the number can be defined in settings.py.


klempine:
> 
> 
> Looks good! 👍 I approves but you could change the commit message according to guidelines: https://chris.beams.io/posts/git-commit/ (Added => Add)

Thaks for the tip! I changed the commit accordingly.

PR #71, opened 2019-11-22
OM-118 | Seed development data

pjlampil:
- Added missing factories for the profile
- Added utility functions to generate data
- Added management command to generate data
- Added tests
- Updated the readme


utils/management/commands/seed_data.py
```diff
@@ -0,0 +1,69 @@
...
+            services = generate_services()
+            self.stdout.write("generating groups...")
+            groups = generate_groups_for_services(services=services)
+            self.stdout.write("assigning permissions...")
+            assign_permissions(groups=groups, services=services)
+            self.stdout.write("generating group admins...")
+            generate_group_admins(groups=groups, faker=faker)
+            profile_count = kwargs["profilecount"]
+            self.stdout.write("generating profiles ({})...".format(profile_count))
+            generate_profiles(kwargs["profilecount"], faker=faker)
```
igordavydsson: could use the var above too
```
generate_profiles(profile_count, faker=faker)
```
not that this would make much difference 😅

pjlampil: Makes all the difference in the world, changed 😄

igordavydsson: 👍

PR #72, opened 2019-11-26
Add primary addresses, phones and emails

Frodotus:
- Add primary addresses, phones and emails
- Add batch mutations for addresses, phones and emails
- Separate Create profile from update profile

### DIFF
--- a/profiles/tests/test_profiles_graphql_api.py
+++ b/profiles/tests/test_profiles_graphql_api.py
@@ -2,13 +2,1004 @@ from string import Template
 
 from django.utils.translation import ugettext_lazy as _
 from graphene import relay
+from graphql_relay.node.node import to_global_id
 from guardian.shortcuts import assign_perm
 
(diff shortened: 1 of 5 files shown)


profiles/tests/test_profiles_graphql_api.py
```diff
@@ -2,13 +2,1009 @@
...
+        }
+    }
+
+    mutation = t.substitute(
+        email_id=to_global_id(type="EmailType", id=email.id),
+        email=email_data["email"],
+        email_type=email_data["email_type"],
+        primary="true",
+    )
+    print(mutation)
```
pjlampil: Printing should be removed

profiles/tests/test_profiles_graphql_api.py
```diff
@@ -2,13 +2,1009 @@
...
+    executed = user_gql_client.execute(mutation, context=request)
+    print(executed)
```
pjlampil: Printing should be removed

PR #84, opened 2019-12-20
OM-230 | Add proper error handling

pjlampil:
- Add support for sentry
- Add support for adding error code
- Let most generic exceptions bubble up instead of catching them in resolvers
- Define error codes as constants
- Map exceptions to error codes
- Fix tests


open_city_profile/exceptions.py
```diff
@@ -0,0 +1,9 @@
+from graphql import GraphQLError
+
+
+class ProfileGraphQLError(GraphQLError):
+    """GraphQLError that is not sent to Sentry."""
+
+
+class ServiceAlreadyExistError(ProfileGraphQLError):
```
igordavydsson: `ServiceAlreadyexist_s_Error` 🙂

open_city_profile/tests/conftest.py
```diff
@@ -39,26 +40,26 @@ def superuser():
 
 @pytest.fixture
 def gql_client():
-    gql_client = GraphQLClient(schema)
+    gql_client = GraphQLClient(schema, format_error=GraphQLView.format_error)
```
igordavydsson: you could subclass `GraphQLClient` and bake in your custom format_error in the subclass to avoid specifying it every time. hm, could also have it as a fixture and let other fixtures extend it and override the `.user` attribute. basically, something can be done to avoid this multiple duplicate lines 😅

open_city_profile/views.py
```diff
@@ -0,0 +1,66 @@
...
+            for error in errors:
+                if hasattr(error, "original_error"):
+                    error = error.original_error
+                sentry_sdk.capture_exception(error)
+
+    @staticmethod
+    def format_error(error):
+        def get_error_code(exception):
+            """Get the most specific error code for the exception via superclass"""
+            for exception in exception.mro():
```
igordavydsson: not bad 👍

open_city_profile/exceptions.py
pjlampil: Nice typo and nicely spotted 👍

open_city_profile/tests/conftest.py
pjlampil: Ended up just having a function that returns the formatted client. We need to refactor that conftest anyways in order to make tests quicker and to not do anything unnecessary stuff between tests.

igordavydsson: OK, good enough for the temp solution 👍

open_city_profile/views.py
pjlampil: With heavy heart I threw recursion away 😄

open_city_profile/exceptions.py
igordavydsson: so, are you gonna fix it? 😅

open_city_profile/views.py
igordavydsson: 😭 😢 😿 👋 🚶

youths/schema.py
```diff
@@ -128,20 +122,11 @@ def mutate(self, info, **kwargs):
...
-                profile = Profile.objects.get(user=info.context.user)
-            except Profile.DoesNotExist:
-                raise GraphQLError(_("No profile found, please create one!"))
+            profile = Profile.objects.get(user=info.context.user)
 
-        try:
-            youth_profile = YouthProfile.objects.get(profile=profile)
-        except YouthProfile.DoesNotExist:
-            raise GraphQLError(_("No youth profile found to update."))
+        youth_profile = YouthProfile.objects.get(profile=profile)
```
tuomas777: Hömpelipöm, not sure if I like this letting `DoesNotExist`s to just bubble up here. Possible issues that come to mind:

* When we have an actual bug somewhere which raises `DoesNotExist`, it won't get sent to Sentry

* When a query or mutation involves multiple objects, like here profile and youth profile, we return the same `OBJECT_DOES_NOT_EXIST` so the client is unable to determine which object is actually missing.

services/tests/test_services_graphql_api.py
```diff
@@ -104,6 +103,5 @@ def test_normal_user_cannot_add_service_multiple_times_mutation(rf, user_gql_cli
     # do the mutation again
     executed = user_gql_client.execute(query, context=request)
     assert "errors" in executed
-    assert executed["errors"][0]["message"] == _(
-        "Service already exists for this profile!"
-    )
+    assert "code" in executed["errors"][0]["extensions"]
+    assert executed["errors"][0]["extensions"]["code"] == SERVICE_ALREADY_EXISTS_ERROR
```
tuomas777: As these checks are probably required more in the future as well, one could consider adding a helper for this, something like `assert_error(executed, SERVICE_ALREADY_EXISTS_ERROR)`

open_city_profile/consts.py
```diff
@@ -0,0 +1,7 @@
+# Common errors
+GENERAL_ERROR = "GENERAL_ERROR"
+PERMISSION_DENIED_ERROR = "PERMISSION_DENIED_ERROR"
+OBJECT_DOES_NOT_EXIST_ERROR = "OBJECT_DOES_NOT_EXIST_ERROR"
+
+# Profile specific errors
+SERVICE_ALREADY_EXISTS_ERROR = "SERVICE_ALREADY_EXISTS_ERROR"
```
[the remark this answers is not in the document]
pjlampil: Agree 100%

youths/schema.py
pjlampil: Actually bubbled up exceptions are captured to sentry ATM. And the model is included in the error message (generated by django) and also in the path field. I'm still not sure about which exceptions we should send to sentry and I think I'm going to write a ticket about that. Because I think most of the time we don't want to send `OBJECT_DOES_NOT_EXIST` to sentry since it's most like due to someone playing around with the API or concurrent use of API where someone deletes a record just before someone else tries to read it.

services/tests/test_services_graphql_api.py
pjlampil: True that. Since there are lots of them I'll write a ticket about this.

open_city_profile/views.py
```diff
@@ -0,0 +1,66 @@
...
+                except KeyError:
+                    continue
+
+        try:
+            error_code = get_error_code(error.original_error.__class__)
+        except AttributeError:
+            error_code = GENERAL_ERROR
+        formatted_error = super(GraphQLView, GraphQLView).format_error(error)
+        if error_code and (
+            isinstance(formatted_error, dict) and "extensions" not in formatted_error
```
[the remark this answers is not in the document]
pjlampil: Extensions might be there because you can throw a GraphQLError and give it `extensions` parameter which would be set here. I think this is useful in cases where you wan't to add all kind of stuff in the extensions (as in spec: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/DD/pages/59965445/GraphQL+guidelines#GraphQL-error-message-format) and cannot rely solely to the exception/error message mapping.

I still think that the if-clause was lacking checking if the actual `code` exists. That's now fixed.

youths/schema.py
tuomas777: Yeah they are captured ATM, my bad. But as you said, most `OBJECT_DOES_NOT_EXIST`s should not be sent to Sentry, so once that is fixed, we'll get the problem I was talking about :)

I think that whatever a user throws at the API, it should not generate a Sentry error. Only bugs should. This probably means that in our code whenever we have a situation where we know a `DoesNotExist` can be raised, we should handle it somehow to prevent it from ending up in Sentry. And then unexcepted `DoesNotExist`s (bugs) will end up in Sentry.

And yes, Django adds the model in the message, but as the client may need to behave differently based on which object is missing, I don't think it's nice that the client then needs to parse the message, which kind of defeats the whole purpose of these error codes :)

pjlampil: This is still the MVP which has been on my desk for too long due to A, B, C ... X, Y & Z. I'd like to improve this in the future since it's absolutely certain that we can't come up with a perfect solution.

The original goal for this story had nothing to do with sentry but it came along during the work. We need to have good guidelines about what to throw at sentry before we start implementing the logic for it.

The only option to keep `DoesNotExist`s away from sentry is definitely not to try-catch them all over the place which just pollutes the code.

I think that instead of using generic `OBJECT_DOES_NOT_EXIST` we could use more accurate error code mappings such as:
```
{
   Profile.DoesNotExist: "PROFILE_DOES_NOT_EXIST_ERROR",
   YouthProfile.DoesNotExist: "YOUTH_PROFILE_DOES_NOT_EXIST_ERROR",
   ...
}
```

open_city_profile/views.py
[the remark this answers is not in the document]
tuomas777: Ok, and now with the code check it makes sense, as it is now possible to add something extra to `extensions` and still get the code there automagically :+1:

I still think it would be nicer if the check was at the start of the method, so it wouldn't figure out the code only to find out later that it is not used for anything, but this is getting nitpicky I guess :)

youths/schema.py
tuomas777: Fair enough :) And that seems like a viable solution indeed 👍

PR #87, opened 2020-01-02
OM-248 | Delete profile

pjlampil:
Mutation to delete profile. Prevent deletion if service `BERTH` is active.


profiles/schema.py
```diff
@@ -366,6 +373,39 @@ def mutate(self, info, **kwargs):
         return UpdateProfile(profile=profile)
 
 
+class DeleteProfile(graphene.Mutation):
+    ok = graphene.Boolean()
+
+    class Arguments:
+        id = graphene.ID(required=True)
+
+    def mutate(self, info, **kwargs):
```
igordavydsson: you could probably ensure that there is a user in the context with `@login_required`

profiles/schema.py
```diff
@@ -366,6 +373,39 @@ def mutate(self, info, **kwargs):
         return UpdateProfile(profile=profile)
 
 
+class DeleteProfile(graphene.Mutation):
+    ok = graphene.Boolean()
+
+    class Arguments:
+        id = graphene.ID(required=True)
```
tuomas777: Is there a reason why this delete needs ID argument but the update does not? To me it seems a bit inconsistent.

pjlampil: IMO using id instead of relying on user is better way to accomplish this and we have discussed and kinda agreed this before with people working on the project. Using id makes the same API usable for admin users and in next PR there will be actually use case for profiles that have no user.

Update should be refactored at some point but there's been too much of that lately so it's going to happen another time. Ticket filed.

profiles/schema.py
```diff
@@ -366,6 +373,39 @@ def mutate(self, info, **kwargs):
...
+        #       https://helsinkisolutionoffice.atlassian.net/browse/OM-248
+        if profile.serviceconnection_set.filter(
+            service__service_type=ServiceType.BERTH
+        ).exists():
+            raise CannotDeleteProfileWhileServiceConnectedError(
+                "Cannot delete profile while service BERTH still connected"
+            )
+
+        result = profile.delete()
+        return DeleteProfile(ok=result)
```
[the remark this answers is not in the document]
pjlampil: Oops. Wrong language/framework. I like your return statement 🥇

profiles/schema.py
pjlampil: Thanks for spotting. I lost it somewhere 😸

profiles/schema.py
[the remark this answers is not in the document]
igordavydsson: would be cool to check for the first element in that tuple, to see if it's `1`. I would expect `ok` to be `true` when it actually does what I expect 😅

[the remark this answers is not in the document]
pjlampil: I was maybe a bit too aggressive with this. Actually we need to check whether it's > 0.

profiles/schema.py
```diff
@@ -366,6 +373,40 @@ def mutate(self, info, **kwargs):
...
+        #       https://helsinkisolutionoffice.atlassian.net/browse/OM-248
+        if profile.serviceconnection_set.filter(
+            service__service_type=ServiceType.BERTH
+        ).exists():
+            raise CannotDeleteProfileWhileServiceConnectedError(
+                "Cannot delete profile while service BERTH still connected"
+            )
+
+        result = profile.delete()
+        return DeleteProfile(ok=result[0] > 0)
```
igordavydsson: maybe:
```python
num_of_deleted_profiles = profile.delete()[0]
return DeleteProfile(ok=num_of_deleted_profiles > 0)
```
or something similar for better readability?

pjlampil: It's not number of deleted profiles, it's number of deleted objects in total (including "sub" objects). IMO Django already ruined the readability of this with the implementation of the return value of delete but what the heck, let's go with that 😄

PR #88, opened 2020-01-03
OM-109 | Add claimable profile query

pjlampil:
Ability for signed in user to use a token to fetch a profile 
that has a valid claim token and doesn't have linked user. 
Profiles can now be created without a linked `User` 
object. By default tokens don't expire but it's possible to 
set an `expires_at` field.

At the moment token is not created automatically for
orphaned profile (however this might be a valid use 
case)


profiles/migrations/0017_auto_20200102_1506.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 2.2.8 on 2020-01-02 13:06
```
igordavydsson: could you rename this migration for the sake of readability?

profiles/models.py
```diff
@@ -67,7 +68,7 @@ def get_notification_context(self):
 
 @reversion.register()
 class Profile(UUIDModel):
-    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    user = models.OneToOneField(User, on_delete=models.CASCADE, null=True, blank=True)
```
igordavydsson: would it make sense to `PROTECT` here? I think `Profile` is more important than `User`  🤔

igordavydsson: but I am open to `CASCADE` as well...

profiles/tests/test_profiles_graphql_api.py
```diff
@@ -1540,3 +1548,81 @@ def test_profile_node_exposes_key_for_federation_gateway(rf, anon_user_gql_clien
...
+    executed = user_gql_client.execute(query, context=request)
+
+    assert "errors" not in executed
+    assert dict(executed["data"]) == expected_data
+
+
+def test_cannot_query_claimable_profile_with_user_already_attached(rf, user_gql_client):
+    profile = ProfileFactory()
+    claim_token = ClaimTokenFactory(
+        profile=profile, expires_at=timezone.now() - timedelta(days=1)
```
igordavydsson: I think you forgot to change some logic here 🙂 right now it's just copypasted from the next test.

profiles/models.py
pjlampil: Maybe `PROTECT` is better for now. This `User`-`Profile` relationship might go through some heavy architecture changes anyways depending on tunnistamo-related stuff.

profiles/tests/test_profiles_graphql_api.py
pjlampil: Actually the accidental copypaste was the `expires_at` argument for the `ClaimTokenFactory`. Otherwise it's correct. The difference is in `ProfileFactory` arguments.

igordavydsson: alrighty 👍

open_city_profile/consts.py
```diff
@@ -2,10 +2,11 @@
...
-PROFILE_DOES_NOT_EXIST_ERROR = "PROFILE_DOES_NOT_EXIST_ERROR"
+TOKEN_EXPIRED_ERROR = "TOKEN_EXPIRED_ERROR"
 
 # Profile specific errors
 SERVICE_CONNECTION_ALREADY_EXISTS_ERROR = "SERVICE_CONNECTION_ALREADY_EXISTS_ERROR"
+PROFILE_DOES_NOT_EXIST_ERROR = "PROFILE_DOES_NOT_EXIST_ERROR"
 CANNOT_DELETE_PROFILE_WHILE_SERVICE_CONNECTED_ERROR = (
     "CANNOT_DELETE_PROFILE_WHILE_SERVICE_CONNECTED_ERROR"
-)
+)
```
igordavydsson: newline missing

open_city_profile/exceptions.py
```diff
@@ -9,9 +9,13 @@ class ServiceAlreadyExistsError(ProfileGraphQLError):
...
+    """Token has expired"""
+
+    
 class CannotDeleteProfileWhileServiceConnectedError(ProfileGraphQLError):
     """Profile cannot be deleted while service is still connected"""
 
 
 class ProfileDoesNotExistError(ProfileGraphQLError):
-    """Profile does not exist"""
+    """Profile does not exist"""
```
igordavydsson: newline missing

pjlampil: Yeah. Tried to be lazy and resolve conflicts with web editor. Never again 😄

PR #91, opened 2020-01-10
OM-295 | Add id argument to mutations

pjlampil:
The purpose of this is to have more generic support for nested mutations
(and to get rid of dependencies where we query `Profile` or `YouthProfile`
via `info.context.user`) where we give youth profile data inside profile data. 
This will enable smoother reuse of the same mutations for administrative
use. We should be able to support following mutations:

- Create new profile and create new youth profile
- Update profile and create new youth profile
- Update profile and update youth profile

In order to accomplish this, following changes have been made:

- Add `id` argument to `updateProfile`
- Add `id` and `profile_id` arguments to `updateYouthProfile`
- Fix `updateProfile` to call correct youth mutation depending on arguments
- Fix permissions
- Fix tests and write additional tests

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -11,7 +11,7 @@ from graphene_django.filter import DjangoFilterConnectionField
 from graphene_django.types import DjangoObjectType
 from graphene_federation import key
 from graphql_jwt.decorators import login_required
-from graphql_relay.node.node import from_global_id
+from graphql_relay.node.node import from_global_id, to_global_id
 from munigeo.models import AdministrativeDivision
 from thesaurus.models import Concept
 
@@ -26,8 +26,9 @@ from services.models import Service, ServiceConnection
 from services.schema import AllowedServiceType, ServiceConnectionType
 from youths.schema import (
     CreateYouthProfile,
+    CreateYouthProfileInput,
     UpdateYouthProfile,
-    YouthProfileFields,
+    UpdateYouthProfileInput,
     YouthProfileType,
 )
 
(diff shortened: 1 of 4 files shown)


profiles/schema.py
```diff
@@ -46,6 +46,56 @@
 )
 
 
+def update_profile(info, profile, profile_data):
+    youth_profile_data = profile_data.pop("youth_profile", None)
+    concepts_of_interest = profile_data.pop("concepts_of_interest", [])
```
igordavydsson: didn't you guys delete these from the `ProfileNode`?

youths/tests/test_youth_profiles_graphql_api.py
```diff
@@ -168,6 +169,44 @@ def test_normal_user_can_create_youth_profile_mutation(rf, user_gql_client):
     assert dict(executed["data"]["createYouthProfile"]) == expected_data
 
 
+def test_superuser_cannot_create_youth_profile_mutation_for_others_profile(
```
igordavydsson: but isn't it supposed to work eventually? I mean, at some point superusers (or just admins with proper permissions) SHOULD be able to do all of these things? so these tests will be removed then? 🙂 and it should happen relatively soon, I guess, as venepaikka would need at least updating profiles by the admins.

profiles/schema.py
pjlampil: Yes, it's already merged in develop. Should I merge those changes into this PR?

youths/tests/test_youth_profiles_graphql_api.py
pjlampil: Maybe I should fix those tests. The purpose was to write tests where we check that no one without either explicit permissions or being the owner of the resource can't do the mutation. The plan was not to remove those test but write new ones for service specific admins. Maybe I should add the support for `serviceType` specific permissions and rewrite the tests since I admit, they are currently a bit misleading.

profiles/schema.py
igordavydsson: yes, please, do 🙂

voneiden:
Closing as stale

PR #93, opened 2020-01-16
OM-337 | My mutations

pjlampil:
Refactor `Profile` and `YouthProfile` mutations to xxxMyMutation to make the separation clearer that those are only for current logged in user to mutate his profile/youth profile. No extra arguments have to be carried for these mutations as the rely heavily on logged in user to fetch all related objects. The approach in this PR is complete opposite of [this](https://github.com/City-of-Helsinki/open-city-profile/pull/91) which should be eventually closed/deleted.

Also: 
- Changed all mutations according to relay spec
- Renamed mutations to avoid name collision with relay spec mutations
- Made transactions in mutations atomic
- Cleaned up mutations of munigeo stuff

### DIFF
--- a/profiles/tests/test_profiles_graphql_api.py
+++ b/profiles/tests/test_profiles_graphql_api.py
@@ -35,11 +35,13 @@ def test_normal_user_can_create_profile(rf, user_gql_client, email_data, profile
         """
             mutation {
                 createMyProfile(
-                    profile: {
-                        nickname: \"${nickname}\",
-                        addEmails:[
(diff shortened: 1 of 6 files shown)


profiles/tests/test_profiles_graphql_api.py
```diff
@@ -284,6 +292,7 @@ def test_normal_user_can_add_address(rf, user_gql_client, address_data):
         """
             mutation {
                 updateMyProfile(
+                    input: {
```
igordavydsson: indent is off now and there is empty line 307 😅

PR #100, opened 2020-01-28
OM-373 | Translate services and add allowed data

pjlampil:
The goal of this PR is to enable frontend to query translated information about connected services which should include title and description of the service and label of each field or data set that is stored into profile and is being read by the service.

Currently we wan't to identify following data sets: Name, email, address, phone and ssn.

- Added translatable `title` and `description` fields to `Service`
- Added model `AllowedDataFields` and made it's field `label` translatable
- Added admin support for both, translations and drag & drop for data field ordering
- Modified schema to support new fields and models


services/tests/test_models.py
```diff
@@ -46,3 +51,10 @@ def test_connect_two_different_services_for_same_profile():
     ServiceConnectionFactory(profile=profile, service=service_1)
     ServiceConnectionFactory(profile=profile, service=service_2)
     assert ServiceConnection.objects.count() == 2
+
+
+def test_allowed_data_fields_get_correct_ordesr():
```
igordavydsson: typo?

services/tests/factories.py
```diff
@@ -19,3 +21,11 @@ class ServiceConnectionFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = ServiceConnection
+
+
+class AllowedDataFieldFactory(factory.django.DjangoModelFactory):
+    field_name = "name"
```
igordavydsson: so, `factory.Faker(...)` did not work here?

services/models.py
```diff
@@ -1,13 +1,42 @@
...
+        return AllowedDataField.objects.all().aggregate(Max("order"))["order__max"] + 1
+    except TypeError:
+        return 1
+
+
+class AllowedDataField(TranslatableModel, SortableMixin):
+    field_name = models.CharField(max_length=30)
+    translations = TranslatedFields(label=models.CharField(max_length=64))
+    order = models.PositiveIntegerField(
+        default=get_next_data_field_order, editable=False, db_index=True
```
igordavydsson: this is not meant to change, right? once you create an object, you cannot change its ~priority?

services/models.py
```diff
@@ -1,13 +1,42 @@
...
+    )
+
+    class Meta:
+        ordering = ["order"]
+
+    def __str__(self):
+        return self.safe_translation_getter("label")
```
igordavydsson: I used to do this:
```python
return self.safe_translation_getter("name", super().__str__())
```
in case you have no translations

services/tests/test_models.py
pjlampil: Most def 🤦‍♂

services/tests/factories.py
pjlampil: It did, but at least for the current test cases it doesn't need to be anything random and actually semantically `field_name` and `label` should stand for the same "thing".

services/models.py
pjlampil: It's meant to change but we don't want to show it on a form. There's a drag & drop UI for changing it:

Screenshot 2020-01-28 at 13 10 26

services/tests/factories.py
igordavydsson: yes, ok, let's go with this code 👍

services/models.py
igordavydsson: ah, ok, that package will handle it. gotcha 👍

PR #101, opened 2020-01-28
Add youth notifications generator and seeder

pjlampil:
Support for generating youth notification templates programatically and added generator to be run in `seed_data`. 

Currently only finnish versions are properly translated. SV & EN are just copies of finnish with `{locale} TRANSLATION NEEDED` text added.

### DIFF
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.contrib.auth.hashers import make_password
 from django.contrib.auth.models import Group
 from django.utils.timezone import get_current_timezone, make_aware
+from django_ilmoitin.models import NotificationTemplate
 from guardian.shortcuts import assign_perm
 
 from profiles.enums import AddressType, EmailType, PhoneType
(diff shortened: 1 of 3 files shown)


utils/utils.py
```diff
@@ -198,3 +199,65 @@ def generate_youth_profiles(percentage=0.2, faker=None):
...
+    template.body_text = fi_text
+    template.set_current_language("sv")
+    template.subject = fi_subject + " SV TRANSLATION NEEDED"
+    template.body_html = fi_html + "<p>SV TRANSLATION NEEDED</p>"
+    template.body_text = fi_text + "<p>SV TRANSLATION NEEDED</p>"
+    template.set_current_language("en")
+    template.subject = fi_subject + " EN TRANSLATION NEEDED"
+    template.body_html = fi_html + "<p>EN TRANSLATION NEEDED</p>"
+    template.body_text = fi_text + "<p>EN TRANSLATION NEEDED</p>"
+    template.save()
```
klempine: Looks like from address is missing, it should be "no-reply@hel.ninja". I think Ville mentioned that the .ninja address should be used when sending stuff to Mailgun.

pjlampil: Comes from the `DEFAULT_FROM_EMAIL` env.

pjlampil: And now it's added to `.gitlab-ci.yml`

PR #103, opened 2020-02-04
OM-397 | Add membership status to youth membership schema

pjlampil:
- Add membership status `enum` to schema
- Resolve membership status based on `expiration` and `approved_time` columns
- Test each case

### DIFF
--- a/youths/tests/test_youth_profiles_graphql_api.py
+++ b/youths/tests/test_youth_profiles_graphql_api.py
@@ -1,6 +1,8 @@
 import uuid
+from datetime import date
 from string import Template
 
+from django.utils import timezone
 from graphql_relay.node.node import to_global_id
 
 from profiles.tests.factories import EmailFactory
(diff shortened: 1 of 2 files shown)


youths/tests/test_youth_profiles_graphql_api.py
```diff
@@ -468,3 +470,57 @@ def test_missing_primary_email_error(rf, youth_profile, anon_user_gql_client):
         executed["errors"][0].get("extensions").get("code")
         == "PROFILE_HAS_NO_PRIMARY_EMAIL_ERROR"
     )
+
+
+def test_pending_youth_profile_should_show_correct_status(rf, user_gql_client):
```
igordavydsson: one test case but parametrized, maybe? since it's like couple of lines of diff

igordavydsson: well, I kinda meant `pytest.mark.parametrize`, seems more like what pytest recommends 😅

pjlampil: Yah. Not really happy about the DRYness of the entire test suite but it's something that needs to be done in a separate task (if ever). 

Decided to go with one test case and 3 different assertions. At least I don't need to repeat the query multiple times now.

igordavydsson: well, okay then 👍

pjlampil: > well, I kinda meant `pytest.mark.parametrize`, seems more like what pytest recommends 😅

I think that suites better for testing pure functions? At least I have to set the state of `YouthProfile`somewhere in order to get particular membership status 🤔

igordavydsson: well, not necessarily. but I was thinking about something like that, before I noticed that you are using different attributes of the `YouthProfile` (`approved_time` and `expiration`), which makes this approach indeed tricky:
```python
@pytest.mark.parametrize("expiration_date, status", [(None, "PENDING"), (date.today(), "EXPIRED"), (timezone.now(), "ACTIVE")])
def test_youth_profile_should_show_correct_membership_status(rf, user_gql_client, expiration_date, status):
    ...
    youth_profile = YouthProfileFactory(profile=profile, expiration=expiration_date)
    ...
    expected_data = {"youthProfile": {"membershipStatus": status}}
```

PR #104, opened 2020-02-05
OM-300 | Add graphiql documentation

klempine:
Added initial GraphiQL descriptions for queries and mutations plus the fields which had the option for adding it. I left the possible error codes mostly as "TODO", since the actual codes must still be checked out. But there's a place now for documenting them at least.

The descriptions have some Markdown formatting in them and they look like this when viewed in a browser:

image


igordavydsson:
Wow, this was pretty nice! 👍 👍 👍

PR #105, opened 2020-02-05
OM-247 | Add download my profile graphql query

pjlampil:
First attempt on serializer that outputs desired models as JSON and serves it through GraphQL endpoint. For more details see comments in `utils/models.py`. Current implementation is somewhat generic but it still need some iterating in order to be able to serve other services as a library (or as a part of the library). The scope of the story/task was to make it work for the profile.

- Serializable mixin
- Serialized all the necessary models
- Simple testing
- Also changed profile-serviceconnections related name that caused some tiny refactors
- Added encryption keys for k8s in order to prevent `SensitiveData` related stuff failing runtime

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,3 +38,4 @@ staging:
         K8S_SECRET_MAIL_MAILGUN_API: "https://api.eu.mailgun.net/v3"
         K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
         K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
+        K8S_SECRET_FIELD_ENCRYPTION_KEYS: "f164ec6bd6fbc4aef5647abc15199da0f9badcc1d2127bde2087ae0d794a9a0b"
--- a/utils/models.py
+++ b/utils/models.py
@@ -1,6 +1,7 @@
 import uuid
 
 from django.db import models
+from django.db.models.fields.reverse_related import OneToOneRel
 
 
 class UUIDModel(models.Model):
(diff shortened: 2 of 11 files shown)


.gitlab-ci.yml
```diff
@@ -38,3 +38,4 @@ staging:
         K8S_SECRET_MAIL_MAILGUN_API: "https://api.eu.mailgun.net/v3"
         K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
         K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
+        K8S_SECRET_FIELD_ENCRYPTION_KEYS: "f164ec6bd6fbc4aef5647abc15199da0f9badcc1d2127bde2087ae0d794a9a0b"
```
igordavydsson: 👀 should it be here?

open_city_profile/models.py
```diff
@@ -0,0 +1,119 @@
+from django.db import models
+
+
+class SerializableMixin(models.Model):
```
igordavydsson: maybe this can be it `utils/models.py`? since it already exists

.gitlab-ci.yml
pjlampil: Probably not. But at the moment it's there just because encrypting something (ssn) without having encryption keys present on runtime will fail (library throws index out of bounds) and before we had a discussion about not having any default values hardcoded in the settings since it's likely that no one remembers to change those for prod. It's not related to the PR but issue was detected during the build of this PR.

open_city_profile/models.py
```diff
@@ -0,0 +1,119 @@
...
+                "accessor": lambda x: x.strftime("%Y-%m-%d")
+            }
+            { "name": "comments" },
+        )
+        serialize_root_key = "POST_OF_THE_DAY"
+
+    class Comment(SerializableMixin):
+        serialize_fields = (
+            { "name": "text" },
+            { "name": "author" },
```
igordavydsson: missing the closing `)` for the tuple 😅

open_city_profile/models.py
pjlampil: Very much true, missed this somehow 😲

utils/models.py
```diff
@@ -8,3 +8,122 @@ class UUIDModel(models.Model):
...
+        serialize_fields = (
+            { "name": "title" },
+            { "name": "content" },
+            {
+                "name": "created_at",
+                "accessor": lambda x: x.strftime("%Y-%m-%d")
+            }
+            { "name": "comments" },
+        )
+        serialize_root_key = "POST_OF_THE_DAY"
```
igordavydsson: is this `serialize_root_key` just for some verbosity? don't understand the technical value behind it

utils/models.py
```diff
@@ -8,3 +8,122 @@ class UUIDModel(models.Model):
...
+    Attributes need to be defined in the extending model:
+
+    - serialize_fields (required)
+        - tuple of dicts:
+            - name (required), name of the field or relation that's going to be added to the serialized object
+            - accessor (optional), function that is called when value of the field is resolved and it takes the
+              actual field value as argument
+
+    - serialize_root_key (required only for the root object)
+        - it's used as key ONLY for the root object (where serializing starts)
```
igordavydsson: is this serialize_root_key just for some verbosity? don't understand the technical value behind it

well, `serialize()` will return a list of the serialized objects and not a dict, but it will still serialize

.gitlab-ci.yml
igordavydsson: ok, well, ideally, I guess that it should use something like [this variable](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/.gitlab-ci.yml#L26), but that could be done later on too

utils/models.py
pjlampil: Excellent point there 🎖. Originally it ended up there basically just because the spec said that `Profile` node needs to have a key with value "CUSTOMER" instead of "PROFILE". I think if we wan't to rename the root key we can do it in the resolver when we are about to resolve the JSON.

I removed the `serialize_root_key`completely and I think it led to a cleaner and more generic solution. I refactored `serialize()` to always return:

```
{
  "key": {model_name_here}
  "children": {fields_here}
}
```

This way now I can just return a list of `obj.serialize()` in manager's serialize method. Only thing I needed to change besides above was to make 1-to-1 relation an exception in order to avoid this

```
{
  "key": "SENSITIVEDATA"
  "children": [
    {
      "key": "SENSITIVEDATA"
      "children": [
        ...
      ]
    }
  ]
}
```

igordavydsson: yeah, looks simpler now 👍

PR #109, opened 2020-02-10
OM-391 | Add sensitive data to schema

pjlampil:
- Make data available in the schema
- Add separate access control for sensitive data


profiles/schema.py
```diff
@@ -201,6 +201,14 @@ class Meta:
         interfaces = (relay.Node,)
 
 
+class SensitiveDataNode(DjangoObjectType):
+    class Meta:
+        model = SensitiveData
+        fields = ("ssn",)
+        filter_fields = []
```
igordavydsson: why does it need an empty `filter_fields` list? 🤔

pjlampil: Good question 🤔. Can't really remember anymore but it had something to do with the Relay. I thought that all the other nodes has that as well but there's actually a couple of those that don't. I think I file a ticket of this and check if the relay filtering works ok without these empty lists. If it does I think each of those entries should be removed.

igordavydsson: if you use `DjangoFilterConnectionField`, then you need to specify which fields are `filter_fields`. otherwise it is not needed.

pjlampil: Oh yeah, that was it 👍 Removing it then.

profiles/schema.py
```diff
@@ -261,6 +273,31 @@ def resolve_phones(self, info, **kwargs):
...
+        #     firstName
+        #     sensitivedata(serviceType: BERTH) {
+        #       ssn
+        #     }
+        #   }
+        # }
+        #
+        # It would be nice to get rid of this requirement sometime in the future.
+
+        service = Service.objects.filter(service_type=kwargs.get("serviceType")).first()
```
igordavydsson: have you investigated the possibility of passing this to the context in those more general resolvers, where you already have this argument. ie. something like this:
```
@staff_required(required_permission="view")
def resolve_profiles(self, info, **kwargs):
    service_type = kwargs.get("serviceType")
    info.context.service_type = service_type
    info.context.save() . # this might not be needed
    return Profile.objects.filter(
        service_connections__service__service_type=service_type
    )
```

then you would be able probably to avoid using the argument here?

igordavydsson: graphene's source code [hints that it should be possible](https://github.com/graphql-python/graphene/blob/8e7d76bbce3eb8755bf4f3beba8b9d8b026c6424/graphene/types/context.py#L14)

pjlampil: I'll give that a try. At first that kind of approach seemed a bit hacky to me but maybe it's worth investigating more.

profiles/schema.py
```diff
@@ -470,6 +495,8 @@ class Query(graphene.ObjectType):
     @staff_required(required_permission="view")
     def resolve_profile(self, info, **kwargs):
         service = Service.objects.get(service_type=kwargs["serviceType"])
+        # serviceType passed on to the sub resolvers
+        info.context.service_type = kwargs["serviceType"]
```
igordavydsson: looks ok, I think 😅

igordavydsson: I definitely prefer this to passing serviceType twice at different query levels 😅 🌞

PR #114, opened 2020-02-19
OM-479 | Fix non-existing 1-to-1 relationships in serializer

pjlampil:
- Refactored serializer to not include non-existenting 1-to-1 relationships in the response
- Test for non-existing sensitive data in graphql query

### DIFF
--- a/utils/models.py
+++ b/utils/models.py
@@ -95,17 +95,20 @@ class SerializableMixin(models.Model):
 
         related_types = {item.name: type(item) for item in model._meta.related_objects}
         if field.get("name") in related_types.keys():
+            value = (
+                getattr(model, field.get("name")).serialize()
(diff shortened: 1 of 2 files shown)


utils/models.py
```diff
@@ -118,6 +121,8 @@ def serialize(self):
         return {
             "key": self._meta.model_name.upper(),
             "children": [
-                self._resolve_field(self, field) for field in self.serialize_fields
+                self._resolve_field(self, field)
+                for field in self.serialize_fields
+                if self._resolve_field(self, field) is not None
```
igordavydsson: just curious: why exactly `is not None`? why not just `if self._resolve_field(self, field)`?

pjlampil: That's an excellent question 🤦‍♂  I guess we can manage without that 👌

PR #116, opened 2020-02-24
OM-479 | Allow photo permissions on create

pjlampil:
- Allow setting `photoUsageApproved` on youth profile create and update for youth over 15 years old
- Return error for under 15 years old if `photoUsageApproved` is set


youths/tests/test_youth_profiles_graphql_api.py
```diff
@@ -282,6 +384,178 @@ def test_normal_user_can_update_youth_profile_mutation(rf, user_gql_client):
...
+        }
+        """
+    )
+    creation_data = {
+        "photoUsageApproved": "true",
+    }
+    query = t.substitute(**creation_data)
+    expected_data = {"youthProfile": {"photoUsageApproved": True}}
+    executed = user_gql_client.execute(query, context=request)
+    print(executed)
```
igordavydsson: 👁

pjlampil: My bad 👀

PR #117, opened 2020-02-26
OM-496 | Add renewable status to schema

pjlampil:
Shows whether youth membership can be renewed

### DIFF
--- a/youths/schema.py
+++ b/youths/schema.py
@@ -57,10 +57,16 @@ class YouthProfileType(DjangoObjectType):
     membership_status = MembershipStatus(
         description="Membership status based on expiration and approved_time fields"
     )
+    renewable = graphene.Boolean(
+        description="Tells if the membership is currently renewable or not"
+    )
 
     class Meta:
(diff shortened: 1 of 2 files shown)


youths/schema.py
```diff
@@ -57,11 +57,17 @@ class YouthProfileType(DjangoObjectType):
...
+    renewable = graphene.Boolean(
+        description="Tells if the membership is currently renewable or not"
+    )
 
     class Meta:
         model = YouthProfile
         exclude = ("id", "approval_token", "language_at_home")
 
+    def resolve_renewable(self, info, **kwargs):
+        return self.expiration != calculate_expiration(date.today())
```
igordavydsson: `!=`?

igordavydsson: ah, I guess not. I think I get it now.

pjlampil: Yeah it's a bit tricky. Maybe some kind of  `is_renewable` helper would have been nice but as it's a one liner I guess that'll do.

PR #118, opened 2020-02-27
VEN-488 | Timmi import admin ui

pjlampil:
Functionality and UI to import customer data (Timmi format) to Profile.
Link to the form can be found under profiles in admin (superuser required).
Upload takes a JSON file with specific format and in case of success returns
a json file that contains the original `customer_id`s and new profile `UUID`s
that were created based on the data. The whole import is atomic that means
either all or none of the profiles are being created.

### DIFF
--- a/templates/admin/profiles/profiles_changelist.html
+++ b/templates/admin/profiles/profiles_changelist.html
@@ -1,0 +1,7 @@
+{% extends 'admin/change_list.html' %}
+
+{% block object-tools %}
+<a href={% url 'admin:upload-json' %}>Import profiles from JSON</a>
+<br />
+{{ block.super }}
+{% endblock %}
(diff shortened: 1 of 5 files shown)


templates/admin/profiles/upload_json.html
```diff
@@ -0,0 +1,20 @@
...
+    </form>
+</div>
+
+<script type="text/javascript"> 
+    document.getElementById("upload-json").addEventListener("submit", function(e) {
+        document.getElementById("submit-button").disabled = true
+    });
+</script>
+
+{% endblock %}
```
igordavydsson: newline missing 👀

templates/admin/profiles/profiles_changelist.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'admin/change_list.html' %}
+
+{% block object-tools %}
+<a href={% url 'admin:upload-json' %}>Import profiles from JSON</a>
+<br />
+{{ block.super }}
+{% endblock %}
```
igordavydsson: newline missing 👀

PR #121, opened 2020-03-04
OM-540 | Timmi connect to berth

pjlampil:
- Add service `BERTH` for imported profiles
- Remove `import Profile` from `Service` to avoid circular imports
- Fix reformats

### DIFF
--- a/profiles/models.py
+++ b/profiles/models.py
@@ -11,6 +11,8 @@ from enumfields import EnumField
 from munigeo.models import AdministrativeDivision
 from thesaurus.models import Concept
 
+from services.enums import ServiceType
+from services.models import Service, ServiceConnection
 from users.models import User
 from utils.models import SerializableMixin, UUIDModel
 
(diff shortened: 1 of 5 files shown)


profiles/models.py
```diff
@@ -179,6 +181,11 @@ def import_customer_data(cls, data):
                     profile.phones.create(
                         phone=phone, phone_type=PhoneType.MOBILE, primary=index == 0
                     )
+                ServiceConnection.objects.create(
+                    profile=profile,
+                    service=Service.objects.get(service_type=ServiceType.BERTH),
```
igordavydsson: Is it a big task to make the `Service` as dropdown in the import form?

igordavydsson: not sure how necessary that is, but I think the import could be used by any service, at least in theory

pjlampil: Not really a huge task, but I would leave it as part of the story where we generalize the whole import. Now this was strictly scoped as `Timmi => Venepaikka` 🤓

igordavydsson: ok, fair enough 👍

PR #122, opened 2020-03-04
OM-466 | Prohibit youth profile for under 13 years old

pjlampil:
- Add validator for `birth_date`
- Add exception and code
- Add test


youths/models.py
```diff
@@ -73,3 +82,7 @@ def __str__(self):
             self.profile.user.last_name,
             self.profile.user.uuid,
         )
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        return super(YouthProfile, self).save(*args, **kwargs)
```
igordavydsson: in python 3 you can just do `super().save(*args, **kwargs)` 🙂

pjlampil: That's very mucho true 🤝

PR #125, opened 2020-03-10
OM-547 | Create youth profile mutation

pjlampil:
- Add `CreateYouthProfileMutation`
- Use `CreateYouthProfileMutation` in `CreateProfileMutation` if there are some youth data present
- Fix `str` method of `YouthProfile` that was failing when no user/profile was present
- Test stuff


youths/schema.py
```diff
@@ -122,6 +126,41 @@ class CreateMyYouthProfileInput(YouthProfileFields):
...
+        youth_profile = CreateMyYouthProfileInput(required=True)
+
+    youth_profile = graphene.Field(YouthProfileType)
+
+    @classmethod
+    @staff_required(required_permission="manage")
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **input):
+        input_data = input.get("youth_profile")
+        if input.get("service_type") != ServiceType.YOUTH_MEMBERSHIP.value:
```
igordavydsson: hm, wouldn't `input.get("service_type")` be `YOUTH_MEMBERSHIP` whereas `ServiceType.YOUTH_MEMBERSHIP.value` is `youth_membership`?

youths/schema.py
```diff
@@ -122,6 +126,41 @@ class CreateMyYouthProfileInput(YouthProfileFields):
...
+            raise CannotPerformThisActionWithGivenServiceType("Incorrect service type")
+
+        profile = Profile.objects.get(pk=from_global_id(input.get("profile_id"))[1])
```
igordavydsson: one trick that I learned is you could do:
```python
profile = graphene.Node.get_node_from_global_id(info, input.get("profile_id"), only_type=ProfileNode)
```
more reliable way to get the object 🙂

youths/schema.py
pjlampil: You could think that, but actually `type(input.get("service_type")) == <class 'str'>` and in this example `input.get("service_type") == 'youth_membership'` 👀

igordavydsson: no, yeah, I expected it to be a string, but I thought it would be all uppercase 🤔 hm, odd. but okay 😅

youths/schema.py
pjlampil: I'm probably not willing to import `ProfileNode` in youth schema because `YouthProfileType` is already imported in profile schema (too circular). There will be huge changes in this when youth profile is going to be separated from profile backend 😄

youths/schema.py
pjlampil:
```
class ServiceType(Enum):
    HKI_MY_DATA = "hki_my_data"
    BERTH = "berth"
    YOUTH_MEMBERSHIP = "youth_membership"
    GODCHILDREN_OF_CULTURE = "godchildren_of_culture"

    class Labels:
        HKI_MY_DATA = _("Helsinki My Data")
        BERTH = _("Berth")
        YOUTH_MEMBERSHIP = _("Youth Membership")
        GODCHILDREN_OF_CULTURE = _("Godchildren of Culture")
```

youths/schema.py
igordavydsson: alright, fair enough 🙂

youths/schema.py
igordavydsson: yeah, I know how the enum is declared. but it's weird that directly getting `input.get("service_type")` does not give you the uppercase string, as graphene's enums are represented. nevermind 🙂 if this works, then this is ok.

PR #128, opened 2020-03-12
OM-548 | Fix postal code field length

pjlampil:
Change postal code db field length from 5 to 32


pjlampil:
@igordavydsson this is for you 😄

PR #129, opened 2020-03-16
DEVOPS-118 | Add production stage to .gitlab-env.yml

ruriky:
Hi,
I would need some help for defining the correct configuration for production environment. The current configuration is copied from the staging env. I tried my best to guess correct config values but some of the are still missing.

All help is appreciated!


pjlampil:
@ruriky Added missing encryption key for production. Not really sure if this is the correct place for it in prod env but we can also sort it out later in order to get the stage up and running.

klempine:
I changed the Sentry DSN directly so that the value is the same as for the staging env.

ruriky:
Thank you! Looks good. The following variables are still unknown but the values will be stored in gitlab (so that they are stored more securely).

"$GL_HEL_STABLE_TUNNISTAMO_OIDC_SECRET"
"$GL_STABLE_DJANGO_SECRET_KEY" (this one I know how to generate)
"$SECRET_STABLE_MAILGUN_API_KEY"

klempine:
> 
> 
> Thank you! Looks good. The following variables are still unknown but the values will be stored in gitlab (so that they are stored more securely).
> 
> "$GL_HEL_STABLE_TUNNISTAMO_OIDC_SECRET"
> "$GL_STABLE_DJANGO_SECRET_KEY" (this one I know how to generate)
> "$SECRET_STABLE_MAILGUN_API_KEY"

I'm not sure if the OIDC_SECRET is needed at all. Maybe you can check if Gitlab has anything stored for the test environment? If not, then it can just be left empty.

The Mailgun API key should be the same as in the test environment. I think I've DM'd you the key in Slack.

.gitlab-ci.yml
```diff
@@ -39,3 +39,24 @@ staging:
...
         K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
         K8S_SECRET_FIELD_ENCRYPTION_KEYS: "f164ec6bd6fbc4aef5647abc15199da0f9badcc1d2127bde2087ae0d794a9a0b"
+
+production:
+    variables:
+        K8S_SECRET_ALLOWED_HOSTS: "*"
+        K8S_SECRET_DEBUG: 0
+        K8S_SECRET_OIDC_SECRET: "$GL_HEL_STABLE_TUNNISTAMO_OIDC_SECRET"
+        K8S_SECRET_SECRET_KEY: "$GL_STABLE_DJANGO_SECRET_KEY"
+        K8S_SECRET_SENTRY_DSN: "https://c2a1cc12588942fd938d8cfdea617cb4:6fc6a7eb191a48bf9be329ca344d6dbc@sentry.hel.ninja/33"
```
tuomas777: While nowadays I think this isn't really considered a strict secret, you can still do at least some harassment with it, so I would not expose it publicly as it is trivial to use a variable instead.

.gitlab-ci.yml
```diff
@@ -39,3 +39,24 @@ staging:
...
+        K8S_SECRET_TOKEN_AUTH_ACCEPTED_SCOPE_PREFIX: "helsinkiprofile"
+        K8S_SECRET_TOKEN_AUTH_AUTHSERVER_URL: "https://api.hel.fi/sso/openid"
+        K8S_SECRET_TOKEN_AUTH_REQUIRE_SCOPE: "True"
+        K8S_SECRET_VERSION: "$CI_COMMIT_SHORT_SHA"
+        K8S_SECRET_MAIL_MAILGUN_KEY: "$SECRET_STABLE_MAILGUN_API_KEY"
+        K8S_SECRET_MAIL_MAILGUN_DOMAIN: "mail.hel.ninja"
+        K8S_SECRET_MAIL_MAILGUN_API: "https://api.eu.mailgun.net/v3"
+        K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
+        K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
+        K8S_SECRET_FIELD_ENCRYPTION_KEYS: "6dda31bbd572d00e6e05c3933b80719770fd6e44645348915c33cf7db1741670"
```
tuomas777: Shouldn't this be a secret?

pjlampil: Probably yes. I have no access to this stuff in Gitlab.

PR #131, opened 2020-03-20
Fix add youth profile via update my profile

pjlampil:
Add possibility to create youth profile via `updateProfile` mutation even if there is no existing youth profile.

### DIFF
--- a/youths/schema.py
+++ b/youths/schema.py
@@ -213,7 +213,9 @@ class UpdateMyYouthProfileMutation(relay.ClientIDMutation):
         )
 
         profile = Profile.objects.get(user=info.context.user)
-        youth_profile, created = YouthProfile.objects.get_or_create(profile=profile)
+        youth_profile, created = YouthProfile.objects.get_or_create(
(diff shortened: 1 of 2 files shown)


youths/schema.py
```diff
@@ -213,7 +213,9 @@ def mutate_and_get_payload(cls, root, info, **input):
         )
 
         profile = Profile.objects.get(user=info.context.user)
-        youth_profile, created = YouthProfile.objects.get_or_create(profile=profile)
+        youth_profile, created = YouthProfile.objects.get_or_create(
```
igordavydsson: so, you will assign the attributes twice? a few lines below:
```python
for field, value in input_data.items():
    setattr(youth_profile, field, value)
```

pjlampil: 🤦‍♂ missed this. I still need to use the defaults youth profile does not exist. What about checking `created`?

igordavydsson: yeah, I think that's what you would need to do

PR #134, opened 2020-03-24
Release/1.0.0 rc.1

tuomas777:


klempine:
Should we do something about Codecov at this point?

pjlampil:
> Should we do something about Codecov at this point?

U mean what?

klempine:
> 
> 
> > Should we do something about Codecov at this point?
> 
> U mean what?

Nvm, just ship it! Ship it real good!

PR #137, opened 2020-03-30
OM-624 | Enable youth registration by an admin

pjlampil:
- Move validation from model to schema
- Test

### DIFF
--- a/youths/schema.py
+++ b/youths/schema.py
@@ -14,6 +14,7 @@ from graphql_relay.node.node import from_global_id
 
 from open_city_profile.exceptions import (
     ApproverEmailCannotBeEmptyForMinorsError,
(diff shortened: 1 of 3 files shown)


youths/schema.py
```diff
@@ -177,6 +178,11 @@ class Input:
     def mutate_and_get_payload(cls, root, info, **input):
         input_data = input.get("youth_profile")
 
+        if calculate_age(input_data["birth_date"]) < 13:
```
igordavydsson: is it okay that the user will be able to modify the birth date later on thru `UpdateMyYouthProfileMutation` and thus her age to below 13 y.o.?

pjlampil: According to current spec, yes 🙄

igordavydsson: well then 👍

PR #138, opened 2020-03-30
OM-572 | Admin UI enhancements

tuomas777:


users/admin.py
```diff
@@ -8,9 +11,15 @@
 
 
 @admin.register(User)
-class UserAdminInline(VersionAdmin):
+class UserAdminInline(VersionAdmin, DjangoUserAdmin):
     inlines = [ProfileAdminInline]
 
+    def get_fieldsets(self, request, obj=None):
+        fieldsets = deepcopy(super().get_fieldsets(request, obj))
+        fieldsets[1][1]["fields"] += ("uuid",)
```
pjlampil: I get this while trying to create a new user:

```
Exception Type: | IndexError
-- | --
tuple index out of range
/app/users/admin.py in get_fieldsets, line 19
```

tuomas777: Yeah, apparently forgot to test user creation with the newest iteration of the thing. Should be fixed now.

pjlampil:
Looks good but how about merging commits to reduce code churn?

tuomas777:
> Looks good but how about merging commits to reduce code churn?

Yup that was the idea, could have been more verbal about that :)

PR #139, opened 2020-04-02
OM-589 | Add notifications to profile

pjlampil:
Added notifications to the profile that can be subscribed by the users. Subscriptions is a separate django app that holds subscription type categories (`SubscriptionTypeCategory`), which are used to group the actual subscription types (`SubscriptionType`), which are basically the actual notifications that users subscribe to. `Subscription` is the model that maps `SubscriptionType` to `Model` and represents the actual subscription. Subscriptions can be managed via GraphQL API and subscriptions can also be manually exported via Django admin in several different formats.

Contents:
- Data model
- GraphQL API
- Django admin with export functionality
- Tests
- Notifications generation


subscriptions/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def get_next_subscription_type_category_order():
+    try:
+        return (
+            SubscriptionTypeCategory.objects.all().aggregate(Max("order"))["order__max"]
+            + 1
+        )
+    except TypeError:
+        return 1
```
tuomas777, line 14: IMO this seems a bit cryptic at first glance, maybe like

```
order_max = SubscriptionTypeCategory.objects.aggregate(Max("order"))["order__max"]
return order_max + 1 if order_max else 1
```

would be easier to read and a bit more robust (it won't swallow other TypeErrors) ?

subscriptions/utils.py
```diff
@@ -0,0 +1,93 @@
...
+                        "value": "Olen halukas tekemään vapaaehtoistyötä liittyen kehittyvään tilanteeseen_EN",
+                    },
+                ),
+            },
+        ),
+    },
+)
+
+
+def generate_subscription_types():
```
tuomas777, line 78: This could maybe be inside a transaction?

profiles/migrations/0023_remove_newsletter_fields.py
```diff
@@ -0,0 +1,15 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("profiles", "0022_add_newsletter_options"),
+    ]
+
+    operations = [
+        migrations.RemoveField(model_name="profile", name="newsletters_via_email",),
+        migrations.RemoveField(model_name="profile", name="newsletters_via_sms",),
+    ]
```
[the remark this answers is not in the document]
pjlampil: Yep. Kept them because there were other contributors to the feature branch. They will cancel out when migrations are eventually squashed at some point.

subscriptions/schema.py
```diff
@@ -0,0 +1,77 @@
...
+
+    class Meta:
+        model = SubscriptionTypeCategory
+        interfaces = (relay.Node,)
+        filter_fields = []
+
+
+class SubscriptionInputType(graphene.InputObjectType):
+    subscription_type_id = graphene.ID(require=True)
+    enabled = graphene.Boolean(require=True)
```
tuomas777, line 39: Hmm, not sure why this isn't failing somehow, but isn't that supposed to be `required` ? 🙂

pjlampil: Good catch. I have no idea why I had that same typo in 2 different files 😑 👍

profiles/schema.py
```diff
@@ -385,7 +392,7 @@ def mutate_and_get_payload(cls, root, info, **input):
 class CreateProfileMutation(relay.ClientIDMutation):
     class Input:
         service_type = graphene.Argument(AllowedServiceType, required=True)
-        profile = ProfileInput(require=True)
+        profile = ProfileInput(requirde=True)
```
tuomas777, line 395: Sorry but I laughed a bit to this fix :D

pjlampil: Hesus christos 🤣 😆 🤣

PR #141, opened 2020-04-06
OM-670 | Add filtering profiles by subscriptions

pjlampil:
Just chaining the "built-in" filters won't do the trick because they aren't executed in the same filter. One way to solve this would have been to call `_next_is_sticky` for the queryset but it would have required maybe too much information about underlying stuff and order of filters. 

Since at least for now the case is always to find subscription types by code and return only profiles that have subscribed to those (`enabled=True`) I found it best option to just have a custom filter for this.

### DIFF
--- a/profiles/tests/test_profiles_graphql_api.py
+++ b/profiles/tests/test_profiles_graphql_api.py
@@ -1754,6 +1754,149 @@ def test_staff_user_can_filter_berth_profiles_by_addresses(rf, user_gql_client):
     assert dict(executed["data"]) == expected_data
 
 
+def test_staff_user_can_filter_berth_profiles_by_subscriptions_and_postal_code(
+    rf, user_gql_client
+):
+    def generate_expected_data(profiles):
+        return {
+            "profiles": {
+                "edges": [
+                    {
+                        "node": {
+                            "emails": {
+                                "edges": [
+                                    {"node": {"email": profile.emails.first().email}}
+                                ]
+                            },
+                            "phones": {
+                                "edges": [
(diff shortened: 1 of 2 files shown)


profiles/tests/test_profiles_graphql_api.py
```diff
@@ -1754,6 +1754,138 @@ def test_staff_user_can_filter_berth_profiles_by_addresses(rf, user_gql_client):
...
+    assert dict(executed["data"]) == generate_expected_data([profile_1, profile_2])
+
+    # test for type 2 + any postal code
+
+    executed = user_gql_client.execute(
+        query,
+        variables={
+            "serviceType": ServiceType.BERTH.name,
+            "subscriptionType": type_2.code,
+            "postalCode": "00",
```
igordavydsson, line 1882: this passes the filter because all profiles have addresses that contain `00` in them, i.e. typical Helsinki addresses? in reality FE client can just skip passing any value to `addresses_PostalCode` filter to actually get any postal code, right?

pjlampil: Yes that's by no means necessary filter. The example use case was just filtering by subscription and postal code. I'll remove the latter postal code filter since it's a bit misleading 🙂

pjlampil: Actually, I'm not going to remove it because it was all about filtering by subscription **and** postal code. I added one more profile and tweaked the expected results a bit.

igordavydsson: yes, that works too 👍

PR #142, opened 2020-04-07
Allow fetching profiles from federation gateway

igordavydsson:
Apollo-federation gateway requires having __resolve_reference
method for all the federated objects across federated backends.

We cannot easily make federation gateway pass serviceType as
query parameter, so we will check if the request user has perms
to view profiles for any of the services connected to the profile
being fetched.

### DIFF
--- a/profiles/utils.py
+++ b/profiles/utils.py
@@ -1,6 +1,12 @@
 import threading
+from typing import TYPE_CHECKING
 
 from graphql_relay.node.node import from_global_id
+
+if TYPE_CHECKING:
+    import profiles.models
+    import users.models
+
 
 _thread_locals = threading.local()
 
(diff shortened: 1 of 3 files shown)


profiles/utils.py
```diff
@@ -47,3 +53,22 @@ def get_current_user():
...
+def user_has_staff_perms_to_view_profile(
+    user: "users.models.User", profile: "profiles.models.Profile"
+) -> bool:
+    """
+    Checks is passed user has "can_view_profiles" permissions
+    for any service connected to the passed profile.
+    """
+
+    service_conns = profile.service_connections.filter(enabled=True)
+    if any(
```
pjlampil, line 67: Could this be just `return any(...`. Don't know about readability, but would save a few LOC 😅

igordavydsson: ah, right, missed that 😅

PR #143, opened 2020-04-07
OM-657 | Add cancel youth membership mutations

pjlampil:
Added mutations for normal users canceling their own youth membership and for admins to cancel an youth membership linked to any profile. Expiration can be set explicitly or it will default today. Also tweaked membership status resolver to calculate status correctly.

### DIFF
--- a/youths/schema.py
+++ b/youths/schema.py
@@ -52,6 +52,15 @@ def create_youth_profile(data, profile):
     return youth_profile
 
 
+def cancel_youth_profile(youth_profile, input):
+    expiration = input.get("expiration")
+
+    youth_profile.expiration = expiration or date.today()
+    youth_profile.save()
+
+    return youth_profile
+
+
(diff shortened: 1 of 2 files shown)


youths/schema.py
```diff
@@ -52,6 +52,18 @@ def create_youth_profile(data, profile):
     return youth_profile
 
 
+def cancel_youth_profile(youth_profile, input):
+    expiration = input.get("expiration", None)
```
tuomas777, line 56: Unnecessary `None`

youths/schema.py
```diff
@@ -52,6 +52,18 @@ def create_youth_profile(data, profile):
...
+
+    if expiration:
+        youth_profile.expiration = expiration
+    else:
+        youth_profile.expiration = date.today()
```
tuomas777, line 61: `youth_profile.expiration = expiration or date.today()` 😉

PR #149, opened 2020-04-23
Add missing configuration into subscriptions' initial migration

charn:
Add missing configuration into the initial migration. Adding a separate migration doesn't seem to cause any SQL to be run, so the change is simply made into the initial migration.

### DIFF
--- a/subscriptions/migrations/0001_initial.py
+++ b/subscriptions/migrations/0001_initial.py
@@ -36,7 +36,7 @@ class Migration(migrations.Migration):
                     ),
                 ),
             ],
-            options={"abstract": False},
+            options={"abstract": False, "ordering": ["order"]},
             bases=(parler.models.TranslatableModelMixin, models.Model),
         ),
         migrations.CreateModel(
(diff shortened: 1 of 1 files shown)


pjlampil:
> Add missing configuration into the initial migration. Adding a separate migration doesn't seem to cause any SQL to be run, so the change is simply made into the initial migration.

Good catch but that's weird behaviour. Is it a bug of Django?

charn:
My first guess would be that this might just be a small human error, like adding the `ordering` meta attribute to the related models after the initial migration was generated and forgetting to regenerate the migration afterwards.

pjlampil:
> Add missing configuration into the initial migration. Adding a separate migration doesn't seem to cause any SQL to be run, so the change is simply made into the initial migration.

Good catch but that's weird behaviour. Is it a bug of Django?

> My first guess would be that this might just be a small human error, like adding the `ordering` meta attribute to the related models after the initial migration was generated and forgetting to regenerate the migration afterwards. 🤷

PR #150, opened 2020-04-23
OM-698 | Make membership number a DB field for YouthProfile

charn:
A DB field for membership number is added to YouthProfile. A post save signal is used for generating a membership number for a new youth profile i.e. re-saving the profile doesn't automatically update the membership number.

DB field is used instead of dynamically calculating the membership number out of the pk of the YouthProfile instance.


youths/signals.py
```diff
@@ -0,0 +1,21 @@
...
+from youths.models import YouthProfile
+
+
+@receiver(post_save, sender=YouthProfile)
+def generate_membership_number(sender, instance, created, **kwargs):
+    """Generate a membership number for a youth profile.
+
+    Membership number is only generated for a new profile.
+    """
+    if created:
```
charn, line 14: Is it a good restriction (right now) to only generate the number for new instances? Also I wonder if a data migration would be needed for existing data?

pjlampil: I think generating this only for new instances is a good restriction. Data migration would be nice 👍

youths/models.py
```diff
@@ -34,6 +34,7 @@ class YouthProfile(models.Model):
     profile = models.OneToOneField(
         Profile, related_name="youth_profile", on_delete=models.CASCADE
     )
+    membership_number = models.CharField(max_length=16, blank=True)
```
pjlampil, line 37: The spec doesn't say anything about it but I'm not sure if wan't allow this to be Blank 🤔

charn: I adjusted the `post_save` signal handler a bit. It will now also generate a membership number when the value is empty. Allowing this to be blank is sort of required, if we would like to be able to create a youth profile though the Django admin, which afaik is the only place currently where the `blank` attribute is considered.

pjlampil:
No wait. Completely forgot the data migration 😑

charn:
Added the datamigration for generating the membership number for existing youth profiles, merging once the tests pass.

PR #151, opened 2020-04-24
Add support for sensitive data

pjlampil:
Add support for sensitive data in `UpdateMyProfile` mutation.

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -101,6 +101,16 @@ def update_profile(profile, profile_data):
         delete_nested(model, profile, data)
 
 
+def update_sensitivedata(profile, sensitive_data):
+    if hasattr(profile, "sensitivedata"):
+        profile_sensitivedata = profile.sensitivedata
+    else:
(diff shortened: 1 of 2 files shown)


profiles/schema.py
```diff
@@ -516,6 +526,7 @@ class Input:
     def mutate_and_get_payload(cls, root, info, **input):
         profile_data = input.pop("profile")
         youth_profile_data = profile_data.pop("youth_profile", None)
+        sensitive_data = profile_data.pop("sensitivedata", None)
```
charn, line 529: Super minor nitpick, on `CreateProfileMutation` this variable is called `sensitivedata`.

pjlampil: No it's a good nitpick 👍. Actually for historical reasons there are some inconsistencies with the naming (`sensitivedata` vs. `sensitive_data`, I'd prefer the latter which would become `sensitiveData` in the API). Maybe I should write a ticket to fix the naming to consistent. It's better to do in a separate ticket because it might break the API.

PR #152, opened 2020-04-24
OM-471 | Primary email for profile

pjlampil:
The idea is to have **one** primary email for each profile which **cannot be blank**. 

In the current implementation `Profile` has 1-to-many emails which have `primary` boolean field. Currently we cannot make sure on the database level that **at least** and **only** one email has the `primary` set to true.

This draft PR has (only) the model and needed migrations for a working solution to solve this problem.

However, there are some cons if we stick to the existing "profile has many emails" structure. Because we have non nullable foreign keys in both ways we need to create new profile instances in a transaction (this could maybe achieved with some creation helper in the profile model?):

```
with transaction.atomic():
    p = Profile()
    e = Email.objects.create(profile=p, email="my@email.com")
    p.primary_email = e
    p.save()
```

Another possible problem that comes to my mind would be how Django admin can handle this data structure. In more detail how could one implement profile create form where it's required to have email added to `email_set` and set as `primary_email` in the same transaction 🤔 

One workaround (at least on the db level) could be still having `primary_email` column as a foreign key in `Profile` and changing emails 1-to-many relation to `additional_emails` relation which would not include the primary_email. We could then have some kind of convenience methods to implement everything that is supported currently.

Feedback appreciated. Do you have some completely different solution in mind? Please feel free to share it 👍


akikoskinen:
I didn't even look at the code but a comment about a future requirement. There will be `Profile`s that don't have any emails! Hence primary email can't be non-null.

voneiden:
Closing as it probably can be closed

PR #155, opened 2020-04-28
OM-508 | Update profile mutation

pjlampil:
- Add update profile mutation
- Add update youth profile mutation (as update profile mutation depends on this)

Known problems intentionally left out of the scope of this task. Filed tickets:
https://helsinkisolutionoffice.atlassian.net/browse/OM-763
https://helsinkisolutionoffice.atlassian.net/browse/OM-764
https://helsinkisolutionoffice.atlassian.net/browse/OM-765
https://helsinkisolutionoffice.atlassian.net/browse/OM-766


profiles/schema.py
```diff
@@ -373,27 +374,53 @@ def __resolve_reference(self, info, **kwargs):
...
+
+class UpdatePhoneInput(PhoneInput):
+    id = graphene.ID(required=True)
+    phone = graphene.String(description="Phone number.")
+    phone_type = AllowedPhoneType(description="Phone number type.")
 
 
 class AddressInput(graphene.InputObjectType):
-    id = graphene.ID()
+    country_code = graphene.String(description="Country code")
```
charn, line 407: Should this description end with a period for consistency?

charn: Though, on the other hand, I think most descriptions might look better without a period, since they aren't proper sentences.

pjlampil: 🤔 Could we file a ticket to make the descriptions & stuff consistent? There are still some work to do (look for `TODO`s 😄) with the API documentation so this is something we could squeeze in there.

profiles/tests/test_profiles_graphql_api.py
```diff
@@ -2317,6 +2318,163 @@ def test_staff_user_cannot_create_a_profile_with_sensitive_data_without_sensitiv
...
+    """
+    )
+    query = t.substitute(
+        service_type=ServiceType.YOUTH_MEMBERSHIP.name,
+        id=to_global_id("ProfileNode", profile.pk),
+        first_name="John",
+    )
+    executed = user_gql_client.execute(query, context=request)
+    assert "errors" in executed
+    assert executed["errors"][0]["message"] == _(
```
charn, line 2473: Probably not necessary, but one extra thing that could be asserted is that the original value in the profile is still there and was not changed.

PR #157, opened 2020-04-29
OM-753 | Make seed_data command re-runnable

charn:
This PR makes data initialization functions re-runnable. Running the functions will not create new objects if there's existing ones.

All environments will run `seed_data` command by default. Production, staging, review environments will run the command when deploying (`on_deploy.sh`). Review environment is also initialized with fake development data. Local development environment/the container itself runs `seed_data` when `APPLY_MIGRATIONS` is set, fake development data is initialized on every startup if `SEED_DEVELOPMENT_DATA` is set (`docker-entrypoint.sh`).

This PR also changes `AllowedDataField.field_name`, `SubscriptionType.code` and `SubscriptionTypeCategory.code` fields to be unique, since they were used in a way that would indicate that they should be so.


klempine:
Looks like you've done a really good job here! Nice catch on the unique constraints also!

I'll let @pjlampil verify this in more detail though 😅

PR #158, opened 2020-04-30
Add flake8 plugin pep8-naming into pre-commit config

charn:
pre-commit config was missing `pep8-naming` plugin for `flake8` which made it miss some errors noticed on Travis.


charn:
Looks like there's some tests that fail when run on the last day of a month. :sweat_smile:

PR #161, opened 2020-05-06
OM-471 | Change profile to have exactly one primary email

pjlampil:
We should make sure that profile has exactly one primary email. With current data model it wasn't possible to create a database constraint for this because a profile has a collection of emails and one of them is marked as primary. 

As a workaround and based on the feedback from [this draft PR](https://github.com/City-of-Helsinki/open-city-profile/pull/152) I decided to validate every API that is able to create or modify profile.

Validation is done in:
- GraphQL API (as mutations are performed inside transaction we can just rollback if validation fails)
- Django admin forms
- Django admin profile import

This also impacted existing tests because of the new rule of profile requiring exactly one primary email.

Also added data migration for existing data to make sure that there are only profiles with *exactly* one primary email.


profiles/models.py
```diff
@@ -203,6 +208,8 @@ def import_customer_data(cls, data):
                     enabled=False,
                 )
                 result[item["customer_id"]] = profile.pk
+            except ProfileMustHaveOnePrimaryEmail as err:
```
charn, line 211: FYI, this can also be written like:

```python
except ProfileMustHaveOnePrimaryEmail:
    raise
```

profiles/models.py
```diff
@@ -262,6 +269,10 @@ class Email(Contact):
     email_type = EnumField(
         EmailType, max_length=32, blank=False, default=EmailType.PERSONAL
     )
+
+    class Meta:
+        ordering = ["-primary"]
```
charn, line 274: Django probably would like to generate a migration for this.

pjlampil:
In addition to suggested fixes added data migration to make existing profiles to have exactly one primary email address.

profiles/migrations/0025_fix_primary_emails.py
```diff
@@ -0,0 +1,38 @@
...
+        if not primary_email:
+            if profile.emails.count() > 0:
+                primary_email = profile.emails.first()
+            else:
+                print(f"Removing profile of {profile.first_name} {profile.last_name}\n")
+                profile.delete()
+                continue
+            primary_email.primary = True
+            primary_email.save()
+        elif primary_emails.count() > 1:
```
charn, line 30: I think you need to make this `elif` an ìf`-clause to make the `4)` on the docstring happen.

pjlampil: I think that `elif` works just fine and saves us one if statement check in 2/4 cases 😄 Wrote a test for that migration as I checked it.

PR #162, opened 2020-05-07
OM-774 | Photo permission related age check is not required for a staff user

charn:
When a staff user is updating youth profile information for a user under 15 years, photo permission related age check is not required.

### DIFF
--- a/youths/tests/test_youth_profiles_graphql_api.py
+++ b/youths/tests/test_youth_profiles_graphql_api.py
@@ -707,6 +707,50 @@ def test_user_cannot_update_youth_profile_with_photo_usage_field_if_under_15_yea
     )
 
 
+def test_staff_user_can_update_youth_profile_with_photo_usage_field_if_under_15_years_old(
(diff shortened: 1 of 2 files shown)


youths/tests/test_youth_profiles_graphql_api.py
```diff
@@ -699,6 +699,50 @@ def test_user_cannot_update_youth_profile_with_photo_usage_field_if_under_15_yea
...
+    service = ServiceFactory(service_type=ServiceType.YOUTH_MEMBERSHIP)
+    group = GroupFactory()
+    user = user_gql_client.user
+    user.groups.add(group)
+    assign_perm("can_manage_profiles", group, service)
+    request = rf.post("/graphql")
+    request.user = user
+
+    # Under 15 years
+    profile = ProfileFactory()
```
pjlampil, line 714: This will fail after base branch update and should be `profile = ProfileWithPrimaryEmailFactory()` 🕵️‍♂️

PR #164, opened 2020-05-11
OM-446 | Get youth data via download api

pjlampil:
Added:
- Download API for youth membership (in future to be separated from profile backend)
- Ability for profile to call service specific backends' download API's and get GDPR data in JSON format
- `gdpr_url` configuration for each service
- Ability to switch GDPR API on/off
- Tests


youths/models.py
```diff
@@ -80,3 +81,17 @@ def __str__(self):
...
+        {"name": "birth_date", "accessor": lambda x: x.strftime("%Y-%m-%d")},
+        {"name": "school_name"},
+        {"name": "school_class"},
+        {"name": "expiration", "accessor": lambda x: x.strftime("%Y-%m-%d")},
+        {"name": "language_at_home", "accessor": lambda x: x.value},
+        {"name": "approver_first_name"},
+        {"name": "approver_last_name"},
+        {"name": "approver_phone"},
+        {"name": "approver_email"},
+        {"name": "expiration", "accessor": lambda x: x.strftime("%Y-%m-%d %H:%M")},
```
charn, line 95: There's two values for expiration.. is that on purpose?

services/models.py
```diff
@@ -38,6 +41,13 @@ class Service(TranslatableModel):
     )
     allowed_data_fields = models.ManyToManyField(AllowedDataField)
     created_at = models.DateTimeField(auto_now_add=True)
+    gdpr_url = models.CharField(
```
charn, line 44: Should we use `URLField` here?

youths/models.py
pjlampil: I don't think so 😃 good catch.

youths/views.py
```diff
@@ -0,0 +1,16 @@
...
+def profiles(request, *args, **kwargs):
+    if settings.GDPR_API_ENABLED:
+        # TODO: Add authentication and security
+        if request.method == "GET":
+            youth_profile = YouthProfile.objects.get(profile__pk=kwargs["id"])
+            return JsonResponse(youth_profile.serialize(), safe=False)
+        else:
+            return HttpResponse(status=405)
+    else:
+        return HttpResponse(status=503)
```
charn, line 16: Would 404 be a more suitable response status, since 5xx usually indicates there's an error happening in the backend?

services/models.py
```diff
@@ -74,3 +84,14 @@ def __str__(self):
...
         {"name": "created_at", "accessor": lambda x: x.strftime("%Y-%m-%d")},
     )
+
+    def download_gdpr_data(self):
+        """Download service specific GDPR data by profile
+        """
+        if self.service.gdpr_url:
+            url = urllib.parse.urljoin(self.service.gdpr_url, str(self.profile.pk))
+            try:
+                return requests.get(url)
```
charn, line 94: You should define a timeout here [since](https://requests.readthedocs.io/en/master/user/quickstart/#timeouts):
> If no timeout is specified explicitly, requests do not time out.

charn: To catch 4xx and 5xx errors (which in `requests` do inherit from `RequestException`) you also need to call `response.raise_for_status()`

```python
response = requests.get(url, timeout=5)
response.raise_for_status()
return response
```

youths/views.py
```diff
@@ -0,0 +1,16 @@
+from django.conf import settings
+from django.http import HttpResponse, JsonResponse
+
+from youths.models import YouthProfile
+
+
+def profiles(request, *args, **kwargs):
+    if settings.GDPR_API_ENABLED:
+        # TODO: Add authentication and security
```
charn, line 9: I guess this endpoint would only be usable by the profile backend? No user client will access this endpoint i.e. adding `login_required` decorator is not what we want here?

pjlampil: I guess we are going to implement some kind of client cert authentication or such. Still to be defined and until then we just disable it.

services/models.py
pjlampil: Check 👍 Thanks for the tips.

services/migrations/0013_change_gdpr_url_to_url_field.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.10 on 2020-05-13 08:19
```
charn, line 1: Redoing the migrations so that only one new migrations is added (`0012_add_gdpr_url.py`) would look nice.

pjlampil: Yeah. Huge merge in between (also had to merge the migrations). Have to do this manually.

services/migrations/0011_add_gdpr_url.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.10 on 2020-05-11 11:16
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("services", "0010_add_sensitivedata_permissions")]
```
charn, line 8: If you rename this file to `0012_add_gdpr_url.py` and change this line to:

`dependencies = [("services", "0011_make_alloweddatafield_field_name_unique")]`

you are able to get rid of the merge migration `services/migrations/0012_merge_20200511_1534.py`

services/models.py
```diff
@@ -74,3 +84,16 @@ def __str__(self):
...
+        """Download service specific GDPR data by profile
+        """
+        if self.service.gdpr_url:
+            url = urllib.parse.urljoin(self.service.gdpr_url, str(self.profile.pk))
+            try:
+                response = requests.get(url, timeout=5)
+                response.raise_for_status()
+                return response
+            except requests.exceptions.RequestException:
+                return response
```
charn, line 98: In case there's an exception should we return `{}` instead? :thinking:

pjlampil: True dat

PR #166, opened 2020-05-14
OM-819 | Change the session cookie setting to use secure cookies

klempine:
Changed the settings so that the server is now using secure cookies. This
issue was brought up by the security auditors and is also recommended to
fix in the Django documentation.


klempine:
As discussed, I changed it so that the secure cookies are enabled only if DEBUG is not enabled.

PR #167, opened 2020-05-19
OM-800 | Validate email format also outside admin

pjlampil:
- Validate email format on save
- Catch validation error in the API and return a GraphQL error
- Test model layer and GraphQL API


profiles/models.py
```diff
@@ -287,6 +287,10 @@ class Meta:
         {"name": "email"},
     )
 
+    def save(self, *args, **kwargs):
+        self.full_clean()
```
charn, line 291: From what I've read/understood there are some arguments against doing a `full_clean` on the save method:
- ModelForm which are used on e.g. the Django admin are already doing this, which leads to calling this method twice.
- It doesn't consider doing only partial saves (only certain fields are being saved)
- It happens before any `pre_save` signals are run, which (in theory) might set values on the model

I don't think any of those reasons are apply enough in the current situation. 

IMHO the best place to do this would be somewhere closer to graphene side of things, but I don't think graphene itself provides any automatic ways of doing this. Though I think it's possible to leverage Django Rest Frameworks serializers and Djangos `ModelForm`s in mutations, but they are probably not really feasible options right now in this PR at least.

Another option for this might be to just call `full_clean` on the mutations. Or just leave this as it is right now. :shrug:

pjlampil: Valid (pun intended) points. Thanks 👍 

I think I had this conversation before with `YouthProfile` and because IMO the only downside is that we are calling `full_clean` twice if we are saving through Django admin forms. Partial saves and doing stuff in `pre_save` signals aren't currently a problemo. And Django admin is still a secondary UI for modifying profile data and used by few.

I prefer calling `full_clean` in save vs. in mutations to keep things DRYer and making it less possible to write a mutation that modifies/creates an email and forgets to validate it.

Introducing DRF serializers or such might be a good idea but probably not in the scope of this task. I'd be interested to know how this is problem tackled in other projects because it's pretty obvious that graphene doesn't do a lot to help with validation.

PR #168, opened 2020-05-19
OM-788 | Make possible to set script name via env

pjlampil:
Make it possible to set `FORCE_SCRIPT_NAME` via envinronment variable. The use case for this is to be able to set `FORCE_SCRIPT_NAME` to `/profiili` in production in order to make it work with `https://api.hel.fi/profiili`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -60,3 +60,4 @@ production:
         K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
         K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
         K8S_SECRET_FIELD_ENCRYPTION_KEYS: "$GL_STABLE_FIELD_ENCRYPTION_KEYS"
+        FORCE_SCRIPT_NAME: "/profiili"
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -60,3 +60,4 @@ production:
         K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
         K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
         K8S_SECRET_FIELD_ENCRYPTION_KEYS: "$GL_STABLE_FIELD_ENCRYPTION_KEYS"
+        FORCE_SCRIPT_NAME: "/profiili"
```
charn, line 63: I'm not totally sure about this, but would we need to add additional `static-map = /profiili/static=/app/static` to `uwsgi.ini` for static files?

pjlampil: @vikoivun speculated the same but then decided that we probably don't need that 👍

vikoivun: I think I've been confused if I said that. As the static files seem to be the thing breaking the graphql introspection tool: https://api.hel.fi/profiili/graphql/
Django should, however, generate correct static file paths, given correct force_script_name setting.

charn: Maybe `MEDIA_URL` `STATIC_URL` settings need to include `FORCE_SCRIPT_NAME` as a prefix? :thinking:

PR #172, opened 2020-05-29
OM-524 | Create a GDPR API for Youth profile data removal

charn:
This PR adds a delete functionality for the GDPR API in Youth profile.

- It also moves youth profile related URL configuration under the `youths` app
- Adds tests for the youth profile GDPR API
- Turns the GDPR API view function into a DRF `APIView`


youths/views.py
```diff
@@ -1,16 +1,48 @@
...
+    dry_run = serializers.BooleanField(required=False, default=False)
+
+
+class YouthProfileGDPRAPIView(APIView):
+    # TODO: Add authentication and security
+
+    def dispatch(self, request, *args, **kwargs):
+        if not settings.GDPR_API_ENABLED:
+            return HttpResponse(status=404)
+        return super().dispatch(request, *args, **kwargs)
```
tuomas777, line 21: If I'm reading this correctly, with this implementation it is not possible for a client to know whether a HTTP 404 response is caused by a non-existing youth profile, or the fact that the whole GDPR API is disabled, which I don't think is optimal.

FWIW the previous way of not including the view to urlpatterns at all when GDPR is disabled seemed like a simpler and more robust way of doing this, but I guess you had some good reason for the change 🙂

youths/tests/test_gdpr_api.py
```diff
@@ -0,0 +1,84 @@
...
+def test_get_profile_information_from_gdpr_api(api_client, youth_profile, snapshot):
+    response = api_client.get(
+        reverse("youths:gdpr", kwargs={"pk": youth_profile.profile.id})
+    )
+
+    assert response.status_code == 200
+    snapshot.assert_match(response.json())
+
+
+@pytest.mark.parametrize("true_value", serializers.BooleanField.TRUE_VALUES)
```
tuomas777, line 30: Just as a matter of principle, I would not use values from DRF but instead values explicitly defined for this test, because now if DRF's TRUE_VALUES change in some version this API will change too, but the test won't catch it. Yes, that will probably never happen in practice :D

youths/views.py
```diff
@@ -1,16 +1,48 @@
...
+        return Response(self.get_object().serialize(), status=status.HTTP_200_OK)
+
+    def delete(self, request, *args, **kwargs):
+        """Delete all data related to the given youth profile.
+
+        Deletes all data related to the given profile id, or just checks if the data can be deleted,
+        depending on the `dry_run` parameter.
+        """
+        if not self.is_dry_run():
+            self.get_object().delete()
```
pjlampil, line 46: If we had some reason why particular youth profile could not be removed how would that be implemented? The initial plan was to use the `dry_run` parameter in such a way that we simulate the deletion without actually committing it. I guess the easiest (and most DRY) way to accomplish this would be doing deletion in transaction and then in the end of the dry run just by calling `transaction.rollback()` to rollback the deletion.

For youth membership I'm not sure will there ever be a situation where the `YouthMembership` model cannot be deleted but at least for Venepaikka there is already use case for that and for that reason it would be nice to already have a placeholder for that logic because I think the idea is to provide with a copy-paste example or probably even a lib solution for other projects to use.

youths/views.py
charn: I removed the setting from urlpatterns mainly because I had some issues with `pytest` and especially when running it with `--doctest-modules` (which is enabled in Travis). When run with that flag, the setting from `conftest` didn't have any effect on the urlpattern. Without that flag, the tests were passing just fine. :shrug: I thought it was less hackier than e.g. reloading the root urlconf for each test. 

I can still try to see if there's a nice way to make this work with having the setting on the urlpattern.

youths/tests/test_gdpr_api.py
charn: I shall consider using explicit values here.

youths/views.py
charn: Though just a side note `get_object_or_404` would have the following message `No YouthProfile matches the given query.`

youths/views.py
charn: I added some transaction related logic into the view. Though maybe for a library like functionality there probably should also be some business logic call to cover cases where `DatabaseError` isn't enough to determine if something can be deleted (e.g. there could a relevant piece of data that is not protected by ForeignKey's `models.PROTECT`).

pjlampil: Agreed, but it doesn't have to be a complete library yet, just keeping our heading towards that. But in general it would be nice to use db constraints as much as possible to avoid writing too many custom checks and codes. Especially having so many different API admin, GraphQL, REST, etc. I'd keep this kind of stuff as low level as possible of course keeping options open for custom checks.

PR #177, opened 2020-06-02
Fix URL and coookie configurations for prod environment

klempine:
Added environment variables and production configurations so that we
can get the production environment working under the path /profiili,
as it's going to be used with api.hel.fi.


open_city_profile/settings.py
```diff
@@ -75,6 +80,22 @@
     SECRET_KEY = "xxx"
 
 ALLOWED_HOSTS = env.list("ALLOWED_HOSTS")
+
+if env.str("CSRF_COOKIE_NAME"):
```
vikoivun, line 84: Note that you don't need to repeat the types here.
So this is enough:
`env("CSRF_COOKIE_NAME")`

.gitlab-ci.yml
```diff
@@ -61,3 +61,10 @@ production:
         K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
         K8S_SECRET_FIELD_ENCRYPTION_KEYS: "$GL_STABLE_FIELD_ENCRYPTION_KEYS"
         K8S_SECRET_FORCE_SCRIPT_NAME: "/profiili"
+        K8S_SECRET_MEDIA_URL: "/profiili/media/"
+        K8S_SECRET_STATIC_URL: "/profiili/static/"
+        K8S_SECRET_CSRF_COOKIE_NAME: "profiili-prod-csrftoken"
+        K8S_SECRET_CSRF_COOKIE_PATH: "/profiili/"
+        K8S_SECRET_CSRF_COOKIE_SECURE: True
+        K8S_SECRET_SESSION_COOKIE_NAME: "profiili-prod-sessionid"
+        K8S_SECRET_SESSION_COOKIE_PATH: "/profiili/"
```
vikoivun, line 70: Secure session cookie is a good idea also.

open_city_profile/settings.py
klempine: I changed the conditionals as per your suggestion.

.gitlab-ci.yml
klempine: I changed the secure session cookie setting so that it can now be set in the environment variables. I also changed the Gitlab config so that test and prod environments are using secure cookies.

PR #179, opened 2020-06-02
Add secret url for printin settings

pjlampil:
"secret" URL endpoint to print the settings


klempine:
So nice that I'm gonna leave a cat image here:

image

PR #184, opened 2020-06-03
OM-420 | Remove user profile and its data from all connected services

charn:
This PR adds a functionality which will remove user profile and data from all the services connected to it.

- Connected services are queried first with a dry run parameter to see if profile's data on that service can be deleted. After an OK from all the connected services, data related to the profile is removed from all the connected services and the profile itself is deleted.
- Data removal URLs for the connected services are configurable


services/models.py
```diff
@@ -45,7 +46,15 @@ class Service(TranslatableModel):
         max_length=2000,
         blank=True,
         help_text=(
-            "Enter the URL of the service. Final URLs are generated by concatenating the url with the profile uuid"
+            "Enter the URL of the service. Final URLs are generated by concatenating the url with the profile UUID."
+        ),
+    )
+    gdpr_removal_url = models.URLField(
```
pjlampil, line 52: At first I thought that it would be good to have only one url that rules them all but maybe it's a nice feature to have an option to have the endpoints in different urls.

charn: I can remove this optional removal URL, since `gdpr_url` will be used for removals as well by default. I just included a way to optionally give a separate removal URL, since I thought it was in the spirit of the ticket OM-351.

pjlampil: Not really sure. @klempine what do you think?

pjlampil: And I also think it depends whether we will have a lot more of these GDPR related urls. We will run out of HTTP methods.

klempine: As discussed in Slack, we can go forward with just one URL per service to keep it simple. We can add more configuration options per service in the future if it's really necessary. I'm not so sure that the GDPR cases would require much more though.

PR #189, opened 2020-06-04
OM-844 | Add missing configurations for review environments

klempine:
The review environments were missing at least the Tunnistamo-related
configurations which probably led to those pesky JWT errors. Let's
see how the environments work with these settings.


klempine:
The Travis build seems to be completed OK but its status is not updating here. I'm just going to force the merging.

PR #192, opened 2020-06-08
Always use JSON output for the YouthProfileGDPRAPIView

charn:
Browseable API templates do not exist, which caused an exception with a normal browser.

Refs OM-420


pjlampil:
Not sure if this is even supposed to work with a browser 😄  Especially after the new authentication is implemented usability will be interesting.

PR #204, opened 2020-07-30
YM-278 | Remove youth profile GQL API

charn:
Remove GraphQL API for youth profile from open-city-profile.

P.S. Doesn't remove models etc. yet to allow them to be migrated.

Refs YM-278


akikoskinen (approved):
👍

PR #206, opened 2020-08-18
OM-837 | Add json format to audit logging

pjlampil:
Changed audit logging to JSON format as in spec: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/416972828/Helsinki+profile+audit+logging


profiles/log_signals.py
```diff
@@ -14,34 +16,73 @@ def should_audit(model):
...
+                },
+            }
+        }
+        service = get_current_service()
+        if service:
+            message["profile_event"]["actor_service"] = {
+                "id": str(service.name),
+                "name": str(service.label),
+            }
+        # print(json.dumps(message))
```
charn, line 68: I think the commented lines here can be removed.

open_city_profile/settings.py
```diff
@@ -282,20 +283,24 @@
...
+    }
+else:
+    audit_handler = {
+        "level": "INFO",
+        "class": "logging.StreamHandler",
+        "stream": stdout,
+    }
 
-AUDIT_LOGGING_ENABLED = env.bool("AUDIT_LOGGING_ENABLED")
+AUDIT_LOGGING_ENABLED = True  # env.bool("AUDIT_LOGGING_ENABLED")
```
charn, line 299: Do we want to enable audit logging by default? If so maybe we could instead change the default value on line 46?

pjlampil: No, we do not. My bad.

open_city_profile/settings.py
```diff
@@ -282,20 +283,24 @@
 # Month from which on the membership will last until the next year, instead of ending in the current year
 YOUTH_MEMBERSHIP_FULL_SEASON_START_MONTH = 5
 
+if not DEBUG:
+    audit_handler = {
+        "level": "INFO",
+        "class": "logging.FileHandler",
+        "filename": "./audit.log",
```
charn, line 290: I think this file will not currently survive when the container is restarted in qa/production environment. Since the logger itself will propagate the log message it probably also ends up in the normal log. I'm just wondering if the file logger is something that we really want here?

pjlampil: Good catch. We want to log to stdout only.

tkutinla:
About log JSON format and content.  Following 2 changes / fixes to do:

#1    "date_time_epoch" value type in JSON should be integer:
- As-is:  "date_time_epoch": "1598258085"
- Should be: "date_time_epoch":  1598258085

#2:   actor.user_id  and target.user_id  should be UUID string values (User's UUID got from authentication), Now they seem to be profile internal values of User object.

Examples:
- As-Is: "target": { ...  "user_id": 8, ......
- Should be:  "target": {... "user_id": "
460bbe7c-d6e2-11ea-a89a-0242ac160005", ....
AND the same in case of actor.user_id field value

profiles/log_signals.py
```diff
@@ -14,34 +16,79 @@ def should_audit(model):
...
+        current_time = datetime.utcnow()
+        current_user = get_current_user()
+        profile = instance.resolve_profile()
+        profile_id = str(profile.pk) if profile else None
+        target_user = profile.user if profile and profile.user else None
+
+        message = {
+            "audit_event": {
+                "origin": "PROFILE-BE",
+                "status": "SUCCESS",
```
charn, line 48: Maybe at some later point of time we'll have to log events with different a status? Just thinking, nothing that needs to be done in this PR. :+1:

charn (approved):
LGTM! :+1: :shipit:

profiles/log_signals.py
pjlampil: Yes that was actually something we already thought. Current implementation doesn't really support for example operations with auth issues 🤔

PR #209, opened 2020-08-27
OM-897 | Fix anonymous user UUID

pjlampil:
Avoid application from crashing in case of user object not having the `uuid` attribute (case `AnonymousUser`)


charn (approved):
LGTM! :+1: :shipit:

PR #212, opened 2020-08-31
OM-903 | Hotfix: Allow deleting youth profile even when its GDPR API is disabled

charn:
Only allow this "override" when GDPR API url hasn't been defined for the youth membership service.


klempine (approved):
LGTM

PR #214, opened 2020-09-02
Clear thread locals after request

pjlampil:
Attempt to fix failing user and service entries in audit log by clearing thread locals after each request


pjlampil:
Seems to be working in review stage. Let's try it out in staging and if everything's still good let's move to production.

klempine (approved):
The street cred on this thing is like 5/5!

charn (approved):
LGTM! :+1: :shipit:

PR #215, opened 2020-09-02
Allow all hosts in docker-compose configuration

charn:
This PR allows all hosts in `docker-compose.yml`. This will allow using the container name as a hostname e.g. in a federation gateway in local development environment.

Refs YM-277


klempine (approved):
LGTM!

PR #216, opened 2020-09-16
YM-180 | Allow changing additional contact persons when approving a youth profile

charn:
This PR adds the possibility to edit additional contact persons when approving a youth profile.


klempine (approved):
Looks good!

PR #222, opened 2020-10-12
Ignore disabled service connection

pjlampil:
Cherry picked only the commit that fixes Venepaikka's issues getting permission denied error if service connection is disabled


klempine (approved):
Blep cat approves: 
image

a-rmz (approved):

PR #224, opened 2020-10-22
Make temporary token generation delete valid token only for current profile

charn:
This PR fixes an issue where generating a temporary token would delete valid tokens for all profiles.

Refs HP-59


akikoskinen (approved):
Super 👍

PR #225, opened 2020-10-27
Dedicated input types for createProfile and updateProfile mutations

akikoskinen:
See commit message for details.

Tested `createProfile` and `updateProfile` mutations after this change using unmodified youth membership admin UI and it worked just fine.

Ref. HP-306

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -525,7 +525,7 @@ class UpdateAddressInput(AddressInput):
     address_type = AllowedAddressType(description="Address type.")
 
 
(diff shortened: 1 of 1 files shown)


akikoskinen:
This change is actually not required by HP-306 anymore, as the design changed so that `createProfile` and `updateProfile` mutations won't be used for this feature. But this PR is still beneficial for the future so this could be merged anyway.

pjlampil (approved):
LGTM 👍

PR #230, opened 2020-11-11
YM-296 | Add management command for exporting youth data

tuomas777:
Added management command `export_youth_data` that creates a JSON file which includes

* `YouthProfile`s and `User`s and `AdditionalContactPerson`s related to those
* `ADGroup`s and `User`s related to the youth admin group `youth_membership`

The JSON file can be consumed by the youth-membership backend's [`import_youth_data`](https://github.com/City-of-Helsinki/youth-membership/pull/15) command. This operation is meant to be executed one time only to migrate data from open-city-profile to youth-membership.

### DIFF
--- a/youths/management/commands/export_youth_data.py
+++ b/youths/management/commands/export_youth_data.py
@@ -1,0 +1,94 @@
+import json
+
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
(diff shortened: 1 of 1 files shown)


charn (approved):
Looks awesome! :+1: :shipit:

PR #231, opened 2020-11-13
Add FIELD_ENCRYPTION_KEYS envvar to review deploys

akikoskinen:
The `FIELD_ENCRYPTION_KEYS` envvar is used by the `django-searchable-encrypted-fields` package. This envvar was already present in the "staging" and "production" deploy configurations.


charn (approved):
LGTM! :+1: :shipit:

PR #234, opened 2020-11-18
YM-278 | Remove youth membership GDPR API view

charn:
View will be available in the new separated youth membership API.

Refs YM-278


akikoskinen (approved):
Good stuff

PR #236, opened 2020-11-20
Profile's verified personal information mutation input validation

akikoskinen:
The string fields are checked for maximum length. There's no more validation than that.

HP-306


akikoskinen:
Changed this PR to a draft. Some more commits will be added...

tituomin (commented):
Looks good. Will wait for more commits before reviewing. Curious whether there will by some custom (domain specific) validation code or only datatype/length validation.

akikoskinen:
> Curious whether there will by some custom (domain specific) validation code or only datatype/length validation.

No plans to do more specific validations. We have no knowledge about how the source data has been validated and I guess we can't really go there and ask them to fix their data, if it doesn't pass some criteria we have set. Based on that the validation here is really lax.

tituomin (approved):
I will accept this for now in order to move things forward.

However, there will be another JIRA task for adding actual positive validation (also specifying what to do when the validation fails). That task will not be specific to this API because validation will have to be added to other endpoints as well.

PR #238, opened 2020-11-27
Update django-helusers and remove allauth dependency

akikoskinen:
Cleaning and updating the authentication system.

### DIFF
--- a/open_city_profile/settings.py
+++ b/open_city_profile/settings.py
@@ -141,7 +141,6 @@ USE_TZ = True
 ENABLE_GRAPHIQL = env("ENABLE_GRAPHIQL")
 
 INSTALLED_APPS = [
-    "helusers.providers.helsinki_oidc",
     "helusers.apps.HelusersConfig",
     "helusers.apps.HelusersAdminConfig",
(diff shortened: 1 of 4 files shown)


open_city_profile/settings.py
```diff
@@ -141,19 +141,15 @@
 ENABLE_GRAPHIQL = env("ENABLE_GRAPHIQL")
 
 INSTALLED_APPS = [
-    "helusers",
-    "helusers.providers.helsinki_oidc",
-    "django.contrib.admin",
+    "helusers.apps.HelusersConfig",
+    "helusers.apps.HelusersAdminConfig",
```
vikoivun, line 145: A curiosity, but it might be possible to leave this out entirely, if we are not using helsinki-tunnistus for authentication into the admin.

vikoivun (approved):
Looks good, I count a net decrease in dependencies.

open_city_profile/settings.py
akikoskinen: Oh. So what do those apps actually do then? I thought they add something into the admin interface.

PR #239, opened 2020-12-09
Remove stale allauth tables and contenttypes

akikoskinen:
In commit 5232d4f213 the dependency to `django-allauth` was removed. But that commit didn't clean allauth stuff away from the database.

I thought it wouldn't be a biggy if the allauth tables are left into the database, but I was wrong. For example `manage.py flush` fails now because it tries to `TRUNCATE` database tables, but doesn't include the allauth tables into the query (because Django doesn't know about these tables anymore). Those allauth tables have foreign key constraints into the other tables that are about to get `TRUNCATE`d, so Postgresql refuses to complete the command at all.

### DIFF
--- a/open_city_profile/apps.py
+++ b/open_city_profile/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class OpenCityProfileConfig(AppConfig):
+    name = "open_city_profile"
+    verbose_name = "Open City Profile"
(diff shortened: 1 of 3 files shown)


pjlampil (commented):
Makes sense to me 🤔

akikoskinen:
Had to add an empty `__init__.py` file in the `open_city_profile` directory. Otherwise `open_city_profile` isn't a proper app. And Django refuses to load it...but only under some circumstances. Development server starts just fine even without that file.

tituomin (approved):

PR #240, opened 2020-12-11
Document that requirements need to be updated within container

akikoskinen:
The `requirements*.txt` files are meant to be used in the containerized environment. Because Python package dependencies are environment dependent, they need to be generated within a container environment.


pjlampil (approved):
Good stuff! 🚢 :shipit:

PR #242, opened 2020-12-15
OM-863 | Store OIDC callback/redirect URI for GDPR API authentication

charn:
This PR adds environment and configuration variable for storing the callback URI related to the GDPR API's OIDC authorization code retrieval. Backend needs this URI when exchanging the authorization code into an access token.

### DIFF
--- a/open_city_profile/settings.py
+++ b/open_city_profile/settings.py
@@ -57,6 +57,7 @@ env = environ.Env(
     USE_X_FORWARDED_HOST=(bool, None),
     CSRF_TRUSTED_ORIGINS=(list, []),
     TEMPORARY_PROFILE_READ_ACCESS_TOKEN_VALIDITY_MINUTES=(int, 2 * 24 * 60),
+    GDPR_AUTH_CALLBACK_URL=(str, ""),
 )
 if os.path.exists(env_file):
     env.read_env(env_file)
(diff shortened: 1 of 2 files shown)


akikoskinen:
This just sets a setting. Where is this setting used? Or is that part coming?

charn:
That part is coming on the next PR.

open_city_profile/settings.py
```diff
@@ -57,6 +57,7 @@
     USE_X_FORWARDED_HOST=(bool, None),
     CSRF_TRUSTED_ORIGINS=(list, []),
     TEMPORARY_PROFILE_READ_ACCESS_TOKEN_VALIDITY_MINUTES=(int, 2 * 24 * 60),
+    GDPR_AUTH_CALLBACK_URL=("", None),
```
akikoskinen, line 60: Shouldn't this be `(str, None)` or perhaps `(str, "")`.

akikoskinen (changes requested):

open_city_profile/settings.py
charn: It should. :sweat_smile:

charn: Okay, fixed!

akikoskinen (approved):

PR #243, opened 2020-12-16
Use GitHub actions for running tests

charn:
This PR migrates the project from Travis into GitHub actions.


akikoskinen (approved):
👍 Seems to do the job.

Was it tested that a red flag is somehow raised if tests don't pass?

charn:
I had some tests failing before I managed to configure GH Actions correctly. Indication for failing tests could be noticed on the PR page.

PR #247, opened 2021-01-07
OM-866 OM-867 | GDPR API uses API tokens when calling connected services

charn:
This PR exchanges the OIDC authorization code received, when GraphQL operations `downloadMyProfile` or `deleteMyProfile` are used, into API tokens related to the connected services of a given profile. These API tokens are used when the backend is doing GDPR related operation on behalf of the user on the connected services.

- Authorization code is exchanged into API tokens
- Correct API token is used for accessing GDPR API of a service


open_city_profile/oidc.py
```diff
@@ -18,3 +27,67 @@ def authenticate(self, request, **kwargs):
             return None
         user, auth = user_auth_tuple
         return user
+
+
+class TunnistamoTokenExchange:
```
akikoskinen, line 32: I'm wondering if this is something that `django-helusers` can/should handle 🤔 [There's already some code](https://github.com/City-of-Helsinki/django-helusers/blob/3f57e79630302401765f4d5a566ca4e8361877e2/helusers/pipeline.py#L111-L140) that does something related, although definitely not exactly the same. But should this functionality be offered by `django-helusers`? Is this something that could benefit multiple projects?

open_city_profile/tests/test_oidc_token_exchange.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_authorization_code_exchange_successful(mocker):
+    tte = TunnistamoTokenExchange()
+    token_response = {"scope": "token"}
+    response = MagicMock()
+    response.json.return_value = token_response
+    mocker.patch.object(TunnistamoTokenExchange, "get_oidc_config")
+    mocker.patch.object(OAuth2Session, "fetch_token")
+    mocker.patch.object(OAuth2Session, "get", return_value=response)
```
akikoskinen, line 18: If I read this correctly, it mocks out quite a lot. If all network requests are done with `requests` (the library), there's also another option to only mock out the requests, like e.g. in [these tests](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/open_city_profile/tests/test_graphql_authentication.py#L40-L45).

akikoskinen (changes requested):
I only got through the first three commits so far. But here are some comments.

open_city_profile/oidc.py
```diff
@@ -18,3 +27,67 @@ def authenticate(self, request, **kwargs):
...
+        ):
+            raise ImproperlyConfigured("Required OAuth/OIDC configuration not set.")
+
+        if not settings.GDPR_AUTH_CALLBACK_URL:
+            raise ImproperlyConfigured(
+                "Required Helsinki profile configuration not set."
+            )
+
+    def fetch_api_tokens(self, authorization_code: str) -> dict:
+        """Exchanges the authorization code into a API tokens that can access open-city-profile API."""
```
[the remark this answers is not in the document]
charn: Ah it is. I moved this code youth-membership, so I probably just forgot to update the docstring.

open_city_profile/settings.py
```diff
@@ -322,3 +324,6 @@
 }
 
 GDPR_AUTH_CALLBACK_URL = env("GDPR_AUTH_CALLBACK_URL")
+SOCIAL_AUTH_TUNNISTAMO_KEY = env("OIDC_CLIENT_ID")
+SOCIAL_AUTH_TUNNISTAMO_SECRET = env("OIDC_CLIENT_SECRET")
+SOCIAL_AUTH_TUNNISTAMO_OIDC_ENDPOINT = env("TOKEN_AUTH_AUTHSERVER_URL")
```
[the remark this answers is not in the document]
akikoskinen: I realised that actually the `TUNNISTAMO` part does make sense because these are used in a context that is Tunnistamo specific: fetching api tokens from the `api-tokens` endpoint (which is a Tunnistamo specific implementation). But the `SOCIAL_AUTH` part still feels weird to me.

services/tests/test_models.py
```diff
@@ -55,41 +59,86 @@ def test_allowed_data_fields_get_correct_orders(allowed_data_field_factory):
...
-    assert response.json() == youth_profile_response
+    response = service_connection.download_gdpr_data(api_token="token")
+    assert response == youth_profile_response
 
 
-def test_download_gdpr_data_with_invalid_service(
-    requests_mock, service_factory, profile, service
-):
-    # setup models
+def test_download_gdpr_data_with_invalid_service(requests_mock, profile, service):
```
akikoskinen, line 77: This test name should state how the service is invalid. I can't tell that from the test body.

akikoskinen (changes requested):
Some more comments. Didn't get through this all just yet. Will continue tomorrow.

utils/tests/utils.py
```diff
@@ -0,0 +1,2 @@
+def assert_match_error_code(response, error_code):
+    assert response["errors"][0].get("extensions").get("code") == error_code
```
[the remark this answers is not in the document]
charn: Didn't notice a suitable file already existed, I'll move this.

akikoskinen:
I went through the whole PR now. No further comments.

open_city_profile/oidc.py
akikoskinen: This comment is just playing with an idea. We can merge the implementation in this PR and consider extracting some functionality into `django-helusers` some other time.

open_city_profile/tests/test_oidc_token_exchange.py
akikoskinen: Just realized that some tests use `requests-mock`, some use `responses`. The two libraries are probably solving the exact same problem and it could be possible to use only one of them. But no need to tackle that in this PR.

open_city_profile/settings.py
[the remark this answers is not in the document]
charn: I'll refactor the naming. The `SOCIAL_AUTH` part was there mainly for allowing someone to easily enable Tunnistamo based login into Django admin, but that is not really the intention of this PR so point of having suitable variable names for that in this PR at least.

open_city_profile/oidc.py
```diff
@@ -18,3 +27,81 @@ def authenticate(self, request, **kwargs):
...
+                token_url=oidc_conf["token_endpoint"],
+                code=authorization_code,
+                client_secret=self.client_secret,
+                include_client_id=True,
+                timeout=self.timeout,
+            )
+        except OAuth2Error as exc:
+            raise TokenExchangeError("Failed to obtain an access token.") from exc
+
+        api_tokens_url = urllib.parse.urljoin(self.oidc_endpoint, "/api-tokens")
```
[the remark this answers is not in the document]
charn: Indeed, I already planned to change this into `urllib.parse.urljoin(self.oidc_endpoint, "api-tokens")`, but maybe a totally separate configuration variable might be a good idea.

open_city_profile/oidc.py
charn: Indeed it looks like there is a function in `django-heluser` that accomplishes a similar thing, though it doesn exchange the authorization token, which is required here. But definitely the functionality in `TunnistamoTokenExchange` could be something that could be extracted to `django-helusers` at some point. :+1:

open_city_profile/tests/test_oidc_token_exchange.py
charn: I'll consider that alternative way of mocking the token fetching.

akikoskinen (approved):

PR #248, opened 2021-01-12
Use dataloader to resolve emails

pjlampil:
An example of using dataloader to resolve emails in a more efficient way. 
Source: https://apirobot.me/posts/django-graphql-solving-n-1-problem-using-dataloaders


voneiden:
Duly noted, but closing as stale.

PR #252, opened 2021-01-25
Verified personal information to my profile

akikoskinen:
Add possibility to query verified personal information via the `myProfile` GraphQL operation.


akikoskinen:
@pjlampil You might want to check this out too. Changes `myProfile` query return type (name). Do you see any potential backwards compatibility problems with that?

tkutinla:
Check and modify / improve audit logging related to create / read / update operations of verified personal information.
This applies also to already approved and merged PR "#227 Add verified personal information create/update GraphQL query."

profiles/tests/test_profiles_graphql_api.py
```diff
@@ -3049,6 +3053,162 @@ def test_database_stays_unmodified_when_mutation_is_not_completed(
...
+        received_info = executed["data"]["myProfile"]["verifiedPersonalInformation"]
+        for at in self.ADDRESS_TYPES:
+            received_address = received_info[to_graphql_name(at)]
+            if at == address_type:
+                assert received_address is None
+            else:
+                assert isinstance(received_address, dict)
+
+    @pytest.mark.parametrize("loa", ["substantial", "high"])
+    def test_high_enough_level_of_assurance_is_required(self, loa, rf, user_gql_client):
```
tituomin, line 3187: Naming change: this doesn't test that `is_required` but that `is_allowed_access`
(This test together with the next one together verify `is_required`)

tituomin (changes requested):
Add audit logging.

open_city_profile/tests/conftest.py
```diff
@@ -102,3 +104,13 @@ def gql_schema(rf, anon_user_gql_client):
 def mock_responses():
     with responses.RequestsMock(assert_all_requests_are_fired=False) as mock_resps:
         yield mock_resps
+
+
+def unix_timestamp_now():
+    epoch = datetime(1970, 1, 1, tzinfo=timezone.utc)
+    return int((datetime.now(tz=timezone.utc) - epoch).total_seconds())
```
mikkokeskinen, line 111: Is there a reason datetime.utcnow().timestamp() is not sufficient?

open_city_profile/tests/conftest.py
```diff
@@ -102,3 +104,13 @@ def gql_schema(rf, anon_user_gql_client):
...
+
+
+@pytest.fixture(name="unix_timestamp_now")
+def unix_timestamp_now_fixture():
```
mikkokeskinen, line 115: After not needing unix_timestamp_now function separately I would rename this. For not needing to set the name manually.

profiles/models.py
```diff
@@ -239,16 +239,30 @@ class VerifiedPersonalInformation(ValidateOnSaveModel, NullsToEmptyStringsModel)
...
+        max_length=1024, blank=True, help_text="Last name."
+    )
+    given_name = fields.EncryptedCharField(
+        max_length=1024, blank=True, help_text="The name the person is called with."
+    )
     national_identification_number = fields.EncryptedCharField(
-        max_length=1024, blank=True
+        max_length=1024,
+        blank=True,
+        help_text="Can be social security number or other person identifier.",
```
mikkokeskinen, line 254: I think this should just read National identification number. As "social security number" hasn't been used in decades.

profiles/models.py
```diff
@@ -297,9 +311,19 @@ class VerifiedPersonalInformationPermanentForeignAddress(
 ):
     RELATED_NAME = "permanent_foreign_address"
 
-    street_address = fields.EncryptedCharField(max_length=1024, blank=True)
-    additional_address = fields.EncryptedCharField(max_length=1024, blank=True)
-    country_code = fields.EncryptedCharField(max_length=3, blank=True)
+    street_address = fields.EncryptedCharField(
+        max_length=1024,
+        blank=True,
+        help_text="Street address or whatever is the _first part_ of the address.",
```
mikkokeskinen, line 317: I think admin doesn't support "\_xxx\_" style emphasis. Consider removing the underscores or replacing with `<em>xxx</em>`

profiles/models.py
```diff
@@ -297,9 +311,19 @@ class VerifiedPersonalInformationPermanentForeignAddress(
...
+    )
+    additional_address = fields.EncryptedCharField(
+        max_length=1024,
+        blank=True,
+        help_text="Additional address information, perhaps town, county, state, country etc.",
+    )
+    country_code = fields.EncryptedCharField(
+        max_length=3, blank=True, help_text="An ISO 3166-1 country code."
```
mikkokeskinen, line 325: Could specify if this is supposed to be two or three character country code.

mikkokeskinen (changes requested):
Didn't try the technical functionality and commented only on superficial concerns.

open_city_profile/tests/conftest.py
akikoskinen: `datetime.utcnow()` returns a naive `datetime`. Calling `timestamp()` for naive `datetime`s treats the `datetime` to be local time. That would result in incorrect values (unless local time happens to be UTC).

But `datetime.now(tz=timezone.utc).timestamp()` should actually work.

profiles/models.py
akikoskinen: That could be true. I have only checked how these texts look in the automatically generated GraphQL documentation and that seems to support markdown.

open_city_profile/tests/conftest.py
mikkokeskinen: Ah. `time.time()` seems to be the better choice.

profiles/models.py
akikoskinen: Yes. Originally I made these fields very lax. That's where the documentation comes from. There's also a plan (HP-436) to make these fields more specified and stricter. The documentation needs to be revisited at that point for every field.

profiles/models.py
akikoskinen: Same thing as with the country code. Originally these fields were very "under specified". But they will be much stricter in near future. Perhaps revisit documentation then?

profiles/tests/test_profiles_graphql_api.py
akikoskinen: Fixed in 7502c5b260.

open_city_profile/tests/conftest.py
akikoskinen: Changed in a8d5cd162c.

profiles/models.py
mikkokeskinen: I would still add the text as truthful as possible right away.

open_city_profile/tests/conftest.py
akikoskinen: `time.time()` doesn't necessarily use the unix epoch so it may not return the unix timestamp.

profiles/models.py
akikoskinen: Can do.

profiles/models.py
akikoskinen: With this country code field the situation is a bit more involved. The plan is that it accepts (via the GraphQL API at least) all versions of country codes, alpha-2, alpha-3 and numeric-3. But then a translation is made to the alpha-2 format and that's also the output format. But we are not there yet.

profiles/models.py
akikoskinen: Changed in 0d6bf75947.

akikoskinen:
> Add audit logging.

@tituomin There's HP-509 for that. Before that we also need to complete HP-178. So, leave out of this PR?

tituomin:
> > Add audit logging.
> 
> @tituomin There's HP-509 for that. Before that we also need to complete HP-178. So, leave out of this PR?

Ok.

open_city_profile/tests/conftest.py
```diff
@@ -107,8 +107,7 @@ def mock_responses():
 
 
 def unix_timestamp_now():
-    epoch = datetime(1970, 1, 1, tzinfo=timezone.utc)
-    return int((datetime.now(tz=timezone.utc) - epoch).total_seconds())
+    return int(datetime.now(tz=timezone.utc).timestamp())
```
mikkokeskinen, line 110: Much better than calculating this by ourselves.

Wouldn't this be much clearer if this was just
```python
@pytest.fixture
def unix_timestamp_now():
    return int(datetime.now(tz=timezone.utc).timestamp())
```

Would be more in line with the other fixtures too. As usually the function is the same name as the fixture.

If there is a compelling reason why the timestamp generation should be separate, it could be better to rename the functions. Something like

```python
def get_unix_timestamp_now():
    return int(datetime.now(tz=timezone.utc).timestamp())

@pytest.fixture
def unix_timestamp_now():
    return get_unix_timestamp_now()
```

mikkokeskinen (changes requested):
Good updates! Though I still think the fixture could be better.

open_city_profile/tests/conftest.py
akikoskinen: The unix timestamp provider is also used as a function (in places where fixtures can't be used) so there's a need for both the function and the fixture. Changed as you proposed in 96359d9. I agree it's cleaner this way.

mikkokeskinen (approved):

profiles/tests/test_profiles_graphql_api.py
akikoskinen: @tituomin Are you satisfied with the implemented changes?

tituomin: Well, now that you ask... To me, `is_needed` and `is_required` are synonyms. I would say `high_enough_level_of_access_gains_access`.

(Sorry to waste time on this... It's just that 'is_needed' to me still implies the implication (not high_enough) -> (no_access) , whereas my formulation does not.)

akikoskinen: No problem. Naming is important. Change in 70e26a4a5b.

tituomin (approved):

PR #254, opened 2021-01-27
HP-398 Enable filtering profiles with ids

akikoskinen:
The `profiles` GraphQL query got a new filtering option: `id`. A list of UUIDs can be provided to select only the wanted profiles.

Since a "list of UUIDs" is a type not directly supported by Graphene and/or django_filters, needed to implement some small classes to get the type of the new filtering argument correct.

### DIFF
--- a/open_city_profile/graphene.py
+++ b/open_city_profile/graphene.py
@@ -1,3 +1,11 @@
+import uuid
+
+import graphene
+from django.forms import MultipleChoiceField
+from django_filters import MultipleChoiceFilter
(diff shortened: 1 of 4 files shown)


akikoskinen:
@a-rmz @haavikko @pjlampil I added you as reviewers, for (at least) two things:

1. @a-rmz @haavikko Would this help with your profile search use case?
2. @pjlampil For generally looking at the implementation/API design, does it look sane.

tituomin (approved):

PR #256, opened 2021-01-28
HP-496 Show verified personal information in admin

mikkokeskinen:
Adds VerifiedPersonalInformation model as a Stacked Inline admin in Profile admin. No user has permission to add, change, or delete the values.

Should we consider audit logging here?

### DIFF
--- a/profiles/admin.py
+++ b/profiles/admin.py
@@ -21,6 +21,7 @@ from profiles.models import (
     Profile,
     SensitiveData,
     TemporaryReadAccessToken,
+    VerifiedPersonalInformation,
 )
 from services.admin import ServiceConnectionInline
 from subscriptions.admin import SubscriptionInline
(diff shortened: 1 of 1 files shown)


akikoskinen (commented):
Looks good so far. The verified personal information also contains three addresses (`permanent_address`, `temporary_address`, `permanent_foreign_address`). Those should be visible too.

tkutinla:
As Aki already mentioned, "Verified personal information" as a concept includes also three addresses.

When it comes to audit logging,  discuss with Aki. We discussed about audit logging related to GraphQL queries and mutations of verified information, he knows what is missing from that. The same audit logging implementation (should) apply to both queries / mutations done through GraphQL and through Django-Admin.

mikkokeskinen:
Ah. Didn't notice the addresses.

akikoskinen (approved):
Nice. Squash the last two commits together and it's ready to go.

PR #257, opened 2021-01-28
Can add a service connection when creating or updating profile with verified personal information

akikoskinen:
Added a new `serviceClientId` argument to the `createOrUpdateProfileWithVerifiedPersonalInformation` GraphQL mutation to create (or enable and existing) connection between the profile and service. Unlike the already existing profile creation/updating mutations with the same kind of feature, this doesn't use the `ServiceType` enum because we want to migrate away from that.


services/models.py
```diff
@@ -69,6 +69,13 @@ def __str__(self):
         return self.safe_translation_getter("title", super().__str__())
 
 
+class ServiceClientId(models.Model):
+    service = models.ForeignKey(
+        Service, on_delete=models.CASCADE, related_name="client_ids"
+    )
+    client_id = models.CharField(max_length=256, null=False, blank=False, unique=True)
```
mikkokeskinen, line 76: Wasn't the max length supposed to be 255?

profiles/schema.py
```diff
@@ -804,6 +807,13 @@ def mutate(parent, info, input):
                 if address.is_empty():
                     address.delete()
 
+        service_client_id = input.pop("service_client_id", None)
+        if service_client_id:
+            service = Service.objects.get(client_ids__client_id=service_client_id)
```
mikkokeskinen, line 812: I'm still not familiar with graphene, but does if validate the service_client_id? So that this can never raise DoesNotExist?

services/models.py
akikoskinen: I already had this code before the 255 values came up and it was so close so I didn't change it. I don't think it needs to be that strict here. The main point is that every possible client id fits into this field.

profiles/schema.py
akikoskinen: Not sure which part of the code exactly you meant, but I'll try to clarify.

Graphene ensures that the argument type is according to the GraphQL schema. In the schema the type of the `service_client_id` argument is string. So, if `service_client_id` exists in `input`, it's a string (could be an empty string or `None` too, but those are falsy, so the execution doesn't get inside the `if` block).

If you meant the `service = Service.objects.get(client_ids__client_id=service_client_id)` part, yes, that can raise a `DoesNotExist` exception -- and that's actually expected. There's also a test for that, called `test_giving_non_existing_service_client_id_results_in_object_does_not_exist_error`. So that exception is turned into a `OBJECT_DOES_NOT_EXIST_ERROR` error code in the response.

mikkokeskinen: Ok. That kind of implicit error handling is a bit strange and non-intuitive. Also there are many changes before that point that I guess don't committed to the database because of the transaction?

services/models.py
mikkokeskinen: 255 would be consistent with fields in many other models in the project, so someone could think that 256 is somehow special. But doesn't really matter.

profiles/schema.py
akikoskinen: > Ok. That kind of implicit error handling is a bit strange and non-intuitive.

At first, yes. This same pattern has been used throughout the code base.

> Also there are many changes before that point that I guess don't committed to the database because of the transaction?

Correct. Also a recurring pattern in the code base.

services/models.py
akikoskinen: That seems to be the case. It's a bit weird. Usually these kind of limits are powers of two (I don't know why, perhaps programmers just think in binary). That's also visible in this code base. You can find 32, 64, 128, 1024... and then... **255** 🤯 I don't have an explanation to this discrepancy.

alvaroarenas (approved):

mikkokeskinen (approved):

PR #258, opened 2021-02-01
Add the ability to write audit log to a separate, rotated, file.

vikoivun:
Audit logs are written to a separate file, rotated by Django. The idea is then, that Filebeat will pick logs from this file.
Using a file gives a bit of safety, should something go wrong with Filebeat. If the logs were, for example, transferred using a TCP socket, any logs generated during Filebeat outage would be lost.

### DIFF
--- a/open_city_profile/settings.py
+++ b/open_city_profile/settings.py
@@ -50,6 +50,7 @@ env = environ.Env(
     VERSION=(str, None),
     AUDIT_LOGGING_ENABLED=(bool, False),
     AUDIT_LOG_USERNAME=(bool, False),
+    AUDIT_LOG_FILENAME=(str, ""),
     ENABLE_GRAPHIQL=(bool, False),
     FORCE_SCRIPT_NAME=(str, ""),
     CSRF_COOKIE_NAME=(str, ""),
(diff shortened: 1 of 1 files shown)


open_city_profile/settings.py
```diff
@@ -322,12 +323,30 @@
...
+    "class": "logging.StreamHandler",
+    "stream": stdout,
+}
+_audit_rotating_file_handler = {
+    "level": "INFO",
+    "class": "logging.handlers.RotatingFileHandler",
+    "filename": AUDIT_LOG_FILENAME,
+    "maxBytes": 100000000,  # 100 MiB
+    "backupCount": 1,  # Keep one previous version
+}
```
akikoskinen, line 339: Are these values needed here? Why not just construct the one that is needed within the following if/else construct?

akikoskinen (changes requested):
Can you implement this on top of the _current_ `develop` branch?

open_city_profile/settings.py
vikoivun: Indeed they are not needed. I suppose I was trying to use variable names to document their purposes. And ended up just duplicating the handler name. Bwah. I'll refactor this.

vikoivun:
> Can you implement this on top of the _current_ `develop` branch?

Well. That was embarrasing. Rebased.

open_city_profile/settings.py
```diff
@@ -336,12 +337,29 @@ def _list_to_string(value):
 
 AUDIT_LOGGING_ENABLED = env.bool("AUDIT_LOGGING_ENABLED")
 AUDIT_LOG_USERNAME = env.bool("AUDIT_LOG_USERNAME")
+AUDIT_LOG_FILENAME = env("AUDIT_LOG_FILENAME")
+
+if AUDIT_LOG_FILENAME:
+    # Rotating log file containing only audit logs and
+    # intended for easy consumption by filebeat
```
akikoskinen, line 344: These comments are pretty useless.

open_city_profile/settings.py
```diff
@@ -336,12 +337,29 @@ def _list_to_string(value):
...
+    _audit_log_handler = {
+        "level": "INFO",
+        "class": "logging.handlers.RotatingFileHandler",
+        "filename": AUDIT_LOG_FILENAME,
+        "maxBytes": 100000000,  # 100 MiB
+        "backupCount": 1,  # Keep one previous version
```
akikoskinen, line 350: These comments are pretty useless. You could write the `maxBytes` value as `100_000_000` to improve readability.

akikoskinen (changes requested):

akikoskinen (approved):

PR #259, opened 2021-02-01
Hotfix: Release DataLoaders for improved performance

charn:
Release #250 since it is urgently required in production.


charn:
This was released using a slightly different branching model https://github.com/City-of-Helsinki/open-city-profile/releases/tag/release-1.2.3-hotfix

PR #260, opened 2021-02-03
Add missing migration file

akikoskinen:
Some help texts were added to model fields in d9da86bae1. Although that doesn't change the database schema, Django still considers it as a change needing a migration. Here it is.


mikkokeskinen (approved):
👍

PR #261, opened 2021-02-03
Added Azure DevOps pipeline

Sukriva:
This pipeline extends the base pipeline and defines CI and PR triggers in GitHub

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@
 trigger:
   branches:
     include:
-    - main
+    - develop
     - feature/*
   paths:
     exclude:
@@ -21,7 +21,7 @@ trigger:
 pr:
   branches:
     include:
-    - main
+    - develop
 
 # By default, use self-hosted agents
 pool: Default
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,38 @@
+# File: open-city-profile/azure-pipelines.yml
+
+# Continuous integration (CI) triggers cause a pipeline to run whenever you push 
+# an update to the specified branches or you push specified tags.
+trigger:
+  branches:
+    include:
+    - main
+    - feature/*
+  paths:
+    exclude:
+    - README.md
+
+# Pull request (PR) triggers cause a pipeline to run whenever a pull request is 
+# opened with one of the specified target branches, or when updates are made to 
+# such a pull request.
+#
+# GitHub creates a new ref when a pull request is created. The ref points to a 
+# merge commit, which is the merged code between the source and target branches 
+# of the pull request.
+pr:
+  branches:
+    include:
+    - main
+
+# By default, use self-hosted agents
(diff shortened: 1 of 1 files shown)


akikoskinen:
I don't think this file should be in this repository as it refers private files in other repositories.

Sukriva:
This file extends the pipeline that is in the azure repo.
Here is an example of the way it's supposed to work:
https://dev.azure.com/City-of-Helsinki/devops-toolchain/_git/devops-handbook?path=%2Fget-started-with-devops.md&_a=preview&anchor=pipeline-templates
And also the ticket for it:
https://helsinkiportaali.atlassian.net/browse/ADK-168

akikoskinen:
> This file extends the pipeline that is in the azure repo.

Yes, this tries to extend a pipeline file. But you can't extend pipelines, only templates. 

> Here is an example of the way it's supposed to work:
> https://dev.azure.com/City-of-Helsinki/devops-toolchain/_git/devops-handbook?path=%2Fget-started-with-devops.md&_a=preview&anchor=pipeline-templates

This is a public open source repository. I don't quite like the idea that private hidden things are referenced from public files. I think another architecture needs to be considered.

> And also the ticket for it:
> https://helsinkiportaali.atlassian.net/browse/ADK-168

I have no access to see that. But I'm really interested on what it is that is the goal here. Apparently triggering some pipelines, I guess. But what pipelines? And on what triggers? Because for example, there's no usable pipeline for PRs yet, so what pipelines are tried to get triggered on PRs?

jussiarpalahti:
@akikoskinen So, the idea here is get pipeline on Azure DevOps to trigger when commits happen in profile's repository. Actual pipeline does CI/CD elsewhere. Trigger or PRs is superfluous at the moment and could be removed. Pipeline decides over on Azure's side when to do its thing and when to ignore it (like for specific branches or tags).

Our deployment process isn't as public as our code, hence the private pipeline. Hopefully other methods arise for connecting Github and CI/CD. For now unfortunately files inside repositories is the only way...

akikoskinen:
Ok. Perhaps we can have this kind of trigger file. I would suggest to name it to `helsinki-azure-pipelines.yml` or something like that to indicate that it's Helsinki specific.

Then a few other things:
- PR triggers shouldn't be used (yet), because there's no usable pipeline to handle those.
- `azure-pipelines.yml` in the private repo is on its way to get obsolete. There are replacing pipelines.
- There's already an automatic trigger in place (in Azure DevOps) that reacts on changes in `develop` branch.

...so what is the reason that this file is needed now, again? I don't think it brings anything that isn't already there currently.

Perhaps once there's a usable pipeline for PRs, we need this kind of file for triggering that pipeline.

voneiden:
Closing this as spring cleanup of old stale PR's. 

We intend to do major rework on open-city-profile pipelines in the near future as part of a larger harmonization of different repository pipelines.

PR #262, opened 2021-02-04
Put correct actor into audit log when using GraphQL API

akikoskinen:
See commit message of 640b3920b8 for explanation what was wrong. All the preceding commits are just refactoring to improve test code.

Fixes sub-task HP-194.


open_city_profile/tests/full_stack_tests.py
```diff
@@ -1,6 +1,7 @@
 import uuid
```
mikkokeskinen, line 1: Now that this file gets renamed anyway, could there be a better name for it? As there are no tests in this file, the _tests suffix is a bit misleading.

Also is "full stack" the best term? Could it be "integration" or "component integration"? Or maybe it could be left out altogether.

So something like "graphql_conf", "graphql_call_settings", "graphql_testing_helpers" could be more helpful names? What do you think?

profiles/utils.py
```diff
@@ -65,7 +65,8 @@ def set_current_service(service):
 
 
 def get_current_user():
-    return getattr(_thread_locals, "user", None)
+    req = getattr(_thread_locals, "request", None)
```
mikkokeskinen, line 68: Could the variable name just be "request"?

profiles/tests/test_audit_log.py
```diff
@@ -79,3 +80,23 @@ def test_audit_log_create(user, enable_audit_log, enable_audit_log_username):
...
+):
+    user = profile.user
+    query = """
+        query {
+            myProfile {
+                id
+            }
+        }"""
+
+    with patch_logger("audit", "info") as cm:
```
mikkokeskinen, line 96: patch_logger seems to be going the way of the Dodo.

From patch_logger function doc:
```
Use unittest.assertLogs() if you only need Python 3 support. This
    private API will be removed after Python 2 EOL in 2020 (#27753).
```

mikkokeskinen (commented):
Couple of comments but otherwise good job!

open_city_profile/tests/full_stack_tests.py
akikoskinen: Agree. Maybe "graphql_test_helpers" would be a nice choice.

akikoskinen: Changed to `graphql_test_helpers` in 529a0ca.

profiles/utils.py
akikoskinen: Yes. Changed within 867f60bced.

profiles/tests/test_audit_log.py
akikoskinen: So it seems, good catch. Replaced it with a Pytest log capture solution in 4172aee324.

mikkokeskinen (approved):
Good job!

PR #264, opened 2021-02-08
OM-931 | Increase the buffer-size for uWSGI

charn:
This PR increases the buffer-size for uWSGI which will allow e.g. to include big lists of
AD-group memberships in a Bearer authentication token.

https://uwsgi-docs.readthedocs.io/en/latest/Options.html#buffer-size

Buffer size was chosen to be the same as in [berth-reservations](https://github.com/City-of-Helsinki/berth-reservations)

Refs OM-931


klempine (approved):
Grrrreat!

image

akikoskinen (commented):
32k feels like fine.

PR #265, opened 2021-02-08
Fix audit log timestamps

akikoskinen:
Local timezone was used for the unix timestamp, but it should always be presented in UTC.


mikkokeskinen (approved):
👍

PR #266, opened 2021-02-10
Add caller's IP address to audit log

akikoskinen:
The HTTP request's "X-Forwarded-For" header is used to get the original client's IP address. If there's no such header, then no IP information goes into the audit log.

Note: this hasn't been tested in any deployment yet, so it's unsure if the used infrastructure really provide this header. A validated guess is that it does.

This PR implements sub-task HP-529. As it's the final sub-task of the parent task, this also completes HP-178.

### DIFF
--- a/profiles/utils.py
+++ b/profiles/utils.py
@@ -81,6 +81,9 @@ def get_original_client_ip():
         forwarded_for = request.headers.get("x-forwarded-for", "")
         client_ip = forwarded_for.split(",")[0] or None
 
+        if not client_ip:
+            client_ip = request.META.get("REMOTE_ADDR")
+
     return client_ip
 
 
--- a/profiles/utils.py
+++ b/profiles/utils.py
@@ -1,6 +1,7 @@
 import threading
 from typing import TYPE_CHECKING
 
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from graphql_relay.node.node import from_global_id
 
@@ -78,8 +79,9 @@ def get_original_client_ip():
 
     request = get_current_request()
     if request:
-        forwarded_for = request.headers.get("x-forwarded-for", "")
-        client_ip = forwarded_for.split(",")[0] or None
+        if settings.USE_X_FORWARDED_FOR:
+            forwarded_for = request.headers.get("x-forwarded-for", "")
+            client_ip = forwarded_for.split(",")[0] or None
 
         if not client_ip:
             client_ip = request.META.get("REMOTE_ADDR")
--- a/profiles/utils.py
+++ b/profiles/utils.py
@@ -64,9 +64,18 @@ def set_current_service(service):
     _thread_locals.service = service
 
 
+def get_current_request():
+    return getattr(_thread_locals, "request", None)
+
+
 def get_current_user():
-    request = getattr(_thread_locals, "request", None)
+    request = get_current_request()
     return getattr(request, "user", None) if request else None
+
+
+def get_original_client_ip():
+    request = get_current_request()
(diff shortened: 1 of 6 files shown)


profiles/utils.py
```diff
@@ -64,11 +64,20 @@ def set_current_service(service):
...
+
 def get_current_user():
-    request = getattr(_thread_locals, "request", None)
+    request = get_current_request()
     return getattr(request, "user", None) if request else None
 
 
+def get_original_client_ip():
+    request = get_current_request()
+    return request.headers.get("x-forwarded-for") if request else None
```
mikkokeskinen, line 78: The value of the x-forwarded-for is comma separated list of ip-addresses (if there were many hops). So you need to split it and take the first.

Also in many cases (e.g. when running locally) the x-forwarded-for is not necessarily set. I think you should use REMOTE_ADDR in that case.

Some projects add a middleware that sets the REMOTE_ADDR from the x-forwarded-for if that is set, so that later you can just use REMOTE_ADDR.

mikkokeskinen (changes requested):

profiles/utils.py
mikkokeskinen: Another thing is, that the usage of x-forwarded-for should be configurable in settings. Because if someone is running this project in production without any proxies, the x-forwarded-for should not be trusted.

akikoskinen: > Another thing is, that the usage of x-forwarded-for should be configurable in settings. Because if someone is running this project in production without any proxies, the x-forwarded-for should not be trusted.

The X-Forwarded-For can probably be spoofed whether there are proxies or not. The client can add the header in the original call and proxies just append their IPs into the list.

mikkokeskinen: Of course. But the point is, the first (as they prepend) can be trusted if we know there are proxies in our deployment. But definitely not when there aren't. That's why it has to be configurable.

mikkokeskinen: I deleted my previous comment, I wasn't thinking straight. :)

akikoskinen: No, the first can't be trusted, it's the one that can be provided by the original caller. Proxies _append_ into the end of the list.

```bash
$ curl -s -H "X-Forwarded-For: look_me_mama" https://httpbin.agw.arodevtest.hel.fi/anything?show_env | grep X-Forwarded-For
    "X-Forwarded-For": "look_me_mama,85.76.82.71:39599",
```

mikkokeskinen: The usage of the header should be configured because when there are no proxies in front, we shouldn't let the user  give the IP-address themselves using the header.

In proxy case, we should have a setting to denote the number of proxies we trust and then take the nth from the end. I think.

akikoskinen: The first one _could_ be made trustful, if the edge proxy (or API GW or whatever there is) would **replace** the X-Forwarded-For header, i.e. discard anything that it potentially received from the client. But it looks that at least the API GW tried above doesn't behave like that.

mikkokeskinen: The whole thing is kind of a mess. Because of course the user could use proxies even before the request hits our gateways et al. Then the first IP would be lost.

akikoskinen: I think the conclusion is that there is no 100% correct solution. Perhaps we should settle for a "mostly works" solution.

mikkokeskinen: I agree. That could be: take the first if from the header if the "USE_X_FORWARDED_FOR" (or something) setting is true, else use REMOTE_ADDR.

mikkokeskinen (approved):
👍

PR #267, opened 2021-02-10
Ignore directories starting with .venv

akikoskinen:
In .gitignore and for some tools.

This can be used as a common prefix for Python virtual environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,7 @@ sftp-config.json
 target/
 var/
 venv/
+.venv*/
 .pytest_cache/
 docker-compose.env.yaml
 .vscode/settings.json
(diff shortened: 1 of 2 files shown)


mikkokeskinen:
This is all well and good, but I'd urge you to use pyenv and pyenv-virtualenv. (As mentioned in the City of Helsinki [coding standards](https://dev.hel.fi/coding-standards#backend-code-standards))

When using pyenv virtualenv the virtualenv is not located in the project directory and thus these are not needed.

tituomin (approved):

akikoskinen:
I think no matter what virtual env tool one uses, the virtual environment(s) can be placed outside the project directory. I suppose it's a matter of opinion where they should reside. I can think of pros and cons with either approach.

PR #268, opened 2021-02-12
Improvements to User admin UI

akikoskinen:
Shows more useful data, enhances search and provides easy access to User's Profile.

### DIFF
--- a/users/admin.py
+++ b/users/admin.py
@@ -19,6 +19,7 @@ class UserAdmin(VersionAdmin, DjangoUserAdmin):
         "last_name",
         "is_staff",
     )
+    search_fields = ("uuid", "first_name", "last_name", "email")
     inlines = [ProfileAdminInline]
 
     def get_fieldsets(self, request, obj=None):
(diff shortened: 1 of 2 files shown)


users/admin.py
```diff
@@ -20,16 +21,23 @@ class UserAdmin(VersionAdmin, DjangoUserAdmin):
...
     )
     search_fields = ("uuid", "first_name", "last_name", "email")
-    inlines = [ProfileAdminInline]
 
     def get_fieldsets(self, request, obj=None):
         fieldsets = super().get_fieldsets(request, obj)
         if obj:
             fieldsets = deepcopy(fieldsets)
-            fieldsets[1][1]["fields"] += ("uuid",)
+            fieldsets[1][1]["fields"] += ("uuid", "profile_uuid")
```
mikkokeskinen, line 29: I've usually used something like "get_profile_uuid" to make it clear that this is not an attribute, but a function. Maybe in this case it could be something like "get_profile_uuid_link"?

akikoskinen: Good point.

akikoskinen: Changed.

mikkokeskinen (approved):
Looking good!

PR #270, opened 2021-02-15
Set names to User from verified personal information

akikoskinen:
With the `createOrUpdateProfileWithVerifiedPersonalInformation` GraphQL mutation, if verified first name and/or last name is provided, they are set into the related `User` object as well.


akikoskinen:
This is a wrong approach and shouldn't be merged.

PR #271, opened 2021-02-16
Buffered audit logging

akikoskinen:
The previous implementation potentially caused multiple extra SQL queries to be made. The Profile model's `post_init` signal was attached directly to the creation of the audit log event. The [documentation for that signal](https://docs.djangoproject.com/en/3.1/ref/signals/#post-init) contains a warning:

> For performance reasons, you shouldn’t perform queries in receivers of `pre_init` or `post_init` signals because they would be executed for each instance returned during queryset iteration.

But that’s exactly what the audit logging system did. It requested `profile.user` and that potentially needed to query the database for the `User`.

The audit logging system is changed in a way that `post_init` signal only adds the model instance to be audit logged. The actual audit log event creation happens later. Currently a request middleware does this audit log "flushing".


profiles/audit_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _format_user_data(audit_event, field_name, user):
+    if user:
+        audit_event[field_name]["user_id"] = (
+            str(user.uuid) if hasattr(user, "uuid") else None
+        )
+        if settings.AUDIT_LOG_USERNAME:
+            audit_event[field_name]["user_name"] = (
+                user.username if hasattr(user, "username") else None
```
tituomin, line 48: This is a question also for @vikoivun :
We have an ASVS requirement 7.3.1: encode user supplied data to prevent log injection. @vikoivun can chime in whether this is a realistic problem. (The data flow seems to be: python dict -> json.dumps -> sent to log system).
https://github.com/OWASP/ASVS/blob/master/4.0/en/0x15-V7-Error-Logging.md#v73-log-protection-requirements

If this is not safe here, the encoding should preferably made in a systematic way, not field-by-field.

tituomin: I now realize the username field is system-generated, so this probably is irrevelant.

profiles/audit_log.py
```diff
@@ -0,0 +1,100 @@
+import json
+import logging
+import threading
+from datetime import datetime, timezone
+
+from django.conf import settings
+
+from .utils import get_current_service, get_current_user, get_original_client_ip
+
+_thread_locals = threading.local()
```
tituomin, line 10: You probably have thought this through, but using thread locals seems to be at least rare in a Django context. @mikkokeskinen can also comment, maybe this still is the best option here. Only alternative I can think of is to store the data in the request or response which also seems ugly.

mikkokeskinen: Yes, it is rare. But when we do logging this way, we don't have any variables to rely on. As the logging is done in the \_\_init\_\_ of the model.

I'd say this is just a temporary solution and we should think the whole audit logging implementation again.

profiles/audit_log.py
vikoivun: Still it is a good idea to check how Elasticsearch handles arbitrary data. I'd guess it is pretty tolerant of almost anything. How Kibana then handles stuff is another matter.

akikoskinen: Django puts this restriction to the username: "150 characters or fewer. Letters, digits and @/./+/-/_ only.".

profiles/audit_log.py
akikoskinen: Thread locals were in use in the audit logging system before too, so I followed the convention. Other solutions can of course be considered, but I wouldn't include them in this PR.

mikkokeskinen (approved):
I think this is good or now. Let's tackle the audit logging in depth later.

PR #272, opened 2021-02-17
YM-296 | Set user field to None profiles without user reference on youth profilele export

charn:
This PR fixes a small issue related to youth profile export. Import (on youth-membership project) uses natural keys and using `None` as a natural key causes an error.


tuomas777 (approved):
👍

PR #273, opened 2021-02-18
YM-460 | Temporarily remove a migration which prevents rollback after release

charn:
Current production version uses the tables which are removed in the related migration. This migrations is not strictly needed for the new production release so, we'll temporarily remove it before the release to allow for a possible rollback using Django migrations.

Refs YM-460

P.S. This migration will be re-added after the release has been deemed successful.

### DIFF
--- a/open_city_profile/migrations/0001_remove_allauth_remnants.py
+++ b/open_city_profile/migrations/0001_remove_allauth_remnants.py
@@ -1,27 +1,12 @@
 from django.db import migrations
 
 
-def remove_contenttypes(apps, schema_editor):
-    ContentType = apps.get_model("contenttypes", "ContentType")
-    ContentType.objects.filter(app_label="account").delete()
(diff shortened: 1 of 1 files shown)


akikoskinen (commented):
A few things comes to mind.

1. Could this be targeted to some temporary branch instead of `develop`?
2. If you are going to apply this migration into a database as a noop as proposed in this PR, how will you ever get the effect of the original migration, i.e. removing the obsolete data and tables?

charn:
1. I can prepare a release branch which includes everything else, but not this migration.
2. I was planning to later add a migration which would have the logic which I now removed, since the migration itself doesn't contain any logic which couldn't be run twice (which would happen e.g. in staging)

That said, I'll go with the option 1.

charn:
Production release will happen from branch release/1.3.0 which will make this PR redundant.

PR #274, opened 2021-02-19
Show any available name in admin UI's Users list

akikoskinen:
Commit b5472f7 adds one more feature from HP-521 that wasn't feasible to implement before #271 got in.

Rest of the commits try to show any name for a `User` in the Users list, be it from the User's `Profile`, or from `VerifiedPersonalInformation` as a last resort.

Searching by these names is also possible. The first and last name fields of `VerifiedPersonalInformation` were changed to regular (unencrypted) fields to enable searching.


tituomin (approved):
Great!

PR #276, opened 2021-02-23
Remove odd ServiceServiceType type from GraphQL schema

akikoskinen:
It looks like nobody ever intended this type (and the `serviceType` field in `ServiceNode`) to be present. `ServiceNode` also contains a `type` field of type `ServiceType` which serves the exactly same purpose.


pjlampil (approved):
LGTM

PR #277, opened 2021-02-24
Revert buffered audit logging

akikoskinen:
Reverts #271 (except commit 5f3155f7f5, which just organized code and is useful by itself).

The buffered audit logging introduced more serious problems than it fixed. The previous version caused some performance issues. The buffered version might cause audit logging to fail, when models are deleted, as they might not be present in the database anymore at the time of logging.


charn (approved):
I can confirm that least the GDPR API was broken. LGTM! :+1: :shipit:

mikkokeskinen (approved):
👍

PR #278, opened 2021-02-24
Enable deleting the related User when deleting a Profile

akikoskinen:
The Profile deletion confirmation screen contains a checkbox that can be used to control if the related User should get deleted as well.

Extending the confirmation screen is a bit cumbersome. In this commit the content block part of the original template is copied and a few lines of custom HTML is inserted into it. The inserted part is indicated with comments in the template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ This repository contains `requirements*.in` and corresponding `requirements*.txt
 5. To install Python requirements run:
     * `pip-sync requirements.txt`
 
+**Note:** when updating dependencies, read the [dependency update checklist](docs/dependency-update.adoc) if there's anything you need to pay attention to.
+
 ## Code format
 
 This project uses
--- a/docs/dependency-update.adoc
+++ b/docs/dependency-update.adoc
@@ -1,0 +1,5 @@
+= Checklist for updating dependencies
+
+== Django
+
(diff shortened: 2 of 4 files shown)


tituomin (commented):
Looks good enough, although the template duplication hurts a bit. Curious to hear what @mikkokeskinen has to say.

mikkokeskinen:
We discussed about this earlier and I came into the conclusion that the template copying is the simplest option.

We should start a document that would list all the customizations made to the admin. So that when upgrading Django we would have a clear picture of the files and methods we should look into.

mikkokeskinen (changes requested):
The code is OK by me, but I'd like to see a document which tells that the delete_confirmation.html admin template is copied and should be checked when upgrading Django.

akikoskinen:
> I'd like to see a document which tells that the delete_confirmation.html admin template is copied and should be checked when upgrading Django.

There's some information in the commit message. Any suggestions where to put such documentation? The ideal solution would probably be that there's some document, that can be consulted when dependencies are updated. But that's probably doomed to get outdated, and probably unlikely that anybody doing the updates will consult it anyway.

Just popped into my mind: what if I put an assert somewhere (hopefully close to the current change) that checks the used Django version and if it's different from the expected, it prints useful information about what to check and terminates the server. That couldn't get bypassed. How about it?

mikkokeskinen:
I would just create a file in a docs folder and reference it in the README. Of course after that it is the maintainers responsibility to keep them up-to-date.

akikoskinen:
Something like e89b49aa92?

mikkokeskinen:
Exactly!

mikkokeskinen (approved):
Good work!

PR #281, opened 2021-03-02
Make service type optional for a Service

akikoskinen:
The possible Services was tightly coupled with the `ServiceType` enum in the Python code, which currently has four values. As a result only four Services were possible, having service type values from the `ServiceType` enum. This coupling needs to be broken so that additional Services can be created which don't have any service type (with the connection to the `ServiceType` enum).

In this PR the Service model got a new field: `name`. It serves a similar purpose to the service type but it's a free form string. Eventually it should replace the service type completely.


services/models.py
```diff
@@ -65,6 +68,15 @@ class Meta:
             ("can_view_sensitivedata", "Can view sensitive data"),
         )
 
+    def save(self, *args, **kwargs):
+        # Convenience for saving Services with only service_type and no name.
+        # When service_type is removed from the code base, this should be
+        # removed as well and every Service creation requires a name at that point.
+        if not self.name and self.service_type:
+            self.name = self.service_type.name
```
mikkokeskinen, line 76: Here the name used is ServiceType.name, but in the migration it's ServiceType.value. Should they both be the same?

services/admin.py
```diff
@@ -1,5 +1,6 @@
 from adminsortable.admin import SortableAdmin
 from django.contrib import admin
+from django.utils.translation import ugettext_lazy as _
```
mikkokeskinen, line 3: The prefixed gettext functions will likely be deprecated in the future. I think we should just use "gettext*" without the u everywhere.

mikkokeskinen (changes requested):

services/models.py
akikoskinen: Yes, definitely. Good catch 👍

services/admin.py
akikoskinen: Agree.

mikkokeskinen (approved):
👍

PR #282, opened 2021-03-03
YM-446 | Get client IP from a forwarded header

charn:
Enable USE_X_FORWARDED_FOR in production and staging to have the correct client IP in audit logs. Federation gateway puts the client IP into _X-Forwarded-For_ header (https://github.com/City-of-Helsinki/youth-federation-gateway/pull/13).


akikoskinen (approved):
Seems correct to me, given that this is really what you want to do 😄

PR #283, opened 2021-03-04
Split profiles GraphQL API tests into multiple files

akikoskinen:
All the tests testing the GraphQL API were in one giant file. It had become very hard to navigate and slow to parse by IDEs etc. Split the tests into separate files, one file for each GraphQL operation.

The tests could use some more refactoring too, but I limited this PR to only contain this splitting.


tituomin (approved):

mikkokeskinen (approved):
Good idea

PR #284, opened 2021-03-05
Use newer JWT authentication system in testing environments

akikoskinen:
Expand testing for this new authentication system. If it looks good, eventually it could be possible to get rid of the old system.


charn (approved):
Let's test this! :+1: :shipit:

PR #285, opened 2021-03-08
Determine Service from JWT "azp" claim

akikoskinen:
Some GraphQL operations take some kind of input for determining the Service from which the requester is "coming from". This has been implemented by arguments and input fields in the GraphQL schema. A new way will be used to determine the Service: the "azp" claim of the JWT the requester uses for authentication. Essentially the Service determination is moved to the authentication layer.

There are five affected GraphQL operations:

- query profile
- query profiles
- mutation createProfile
- mutation updateProfile
- mutation addServiceConnection

The old way is still present and working for backwards compatibility. Actually the old way takes precedence if such input is provided, because the old way has enabled doing GraphQL queries that will not be possible with the new way (see commit messages for details about this).


akikoskinen:
FYI @pjlampil @charn @klempine You might be interested to know about this, although shouldn't need any actions from you.

open_city_profile/middleware.py
```diff
@@ -0,0 +1,26 @@
+from django.conf import settings
+from django.core.exceptions import MiddlewareNotUsed
+from helusers.oidc import RequestJWTAuthentication
+
+
+class JWTAuthentication:
```
mikkokeskinen, line 8: Is there a good reason for this being a middleware and not a proper authentication backend?

services/utils.py
```diff
@@ -0,0 +1,19 @@
+from services.models import ServiceClientId
+
+
+def set_service_to_request(request):
```
mikkokeskinen, line 4: Could this be a middleware?

profiles/decorators.py
```diff
@@ -37,18 +37,34 @@ def wrapper(context, *args, **kwargs):
...
+                    "can_{}_profiles".format(required_permission), service
+                ):
+                    return function(*args, **kwargs)
+                else:
+                    raise PermissionDenied(
+                        _("You do not have permission to perform this action.")
+                    )
+            finally:
+                if old_service:
+                    context.service = old_service
```
mikkokeskinen, line 66: Should the "current service" be set here too?

open_city_profile/middleware.py
akikoskinen: The Django authentication backends can only report authenticated/not authenticated, without any reasons why (not authenticated). We want to return some information about the reasons.

Django also mixes authentication with authorization (as if both could be performed at the same point in the request handling), but with GraphQL requests that doesn't really work.

It's possible to make a GraphQL query, where e.g. one part could need an authenticated user and another one not. In such case we always return data for the part that doesn't require authentication, and check the authentication status for the part requiring it and return either data or error (with an appropriate reason) for that part. Authentication and authorization happen in different stages with GraphQL and authorization can even happen multiple times, with different criteria.

services/utils.py
akikoskinen: There are currently two JWT based authentication systems in this project, an [older one](https://github.com/City-of-Helsinki/open-city-profile/blob/HP-575-service-from-azp-claim/open_city_profile/oidc.py#L12-L30) and a [newer one](https://github.com/City-of-Helsinki/open-city-profile/blob/HP-575-service-from-azp-claim/open_city_profile/middleware.py#L8-L29). Only one or the other can be in use, there's a [feature flag](https://github.com/City-of-Helsinki/open-city-profile/blob/HP-575-service-from-azp-claim/open_city_profile/settings.py#L67) to toggle which one.

This helper is called from both of those two authentication systems. Once we get rid of the older authentication system, then this helper could be merged into the newer authentication middleware, if it feels like the right thing to do.

profiles/decorators.py
akikoskinen: Hmm, for a long time I thought that it doesn't have to be, thinking that the request handling is pretty much done and it doesn't make a difference. But I was mistaken. There can off course be other GraphQL fields to be resolved after this decorator exits and the current service should be restored for those. I will fix this.

akikoskinen: Darn, this actually revealed another problem with the audit log implementation. The GraphQL field resolution (which goes through this decorator we are talking about) produces Promises, and those Promises are then settled later. The Promises produce the actual data, query the database via Django models etc. So if "service X" is set as "current" when the field resolution happens, it might not be the "current service" anymore when the Promise is settled and the audit log entry is produced. This can produce wrong "actor service" information into the audit log.

This won't be a problem any more after the `serviceType` arguments -- which allow acting "on behalf of" multiple services within a single GraphQL request -- will get removed (or are ignored) from the GraphQL schema. When "actor service" is determined only from the "azp" claim, there can be only one "actor service" per GraphQL request.

I suppose the GraphQL API was never intended to be used in a way where one can act on behalf of many services, and I doubt any of the users of this API use it that way. Perhaps e.g @charn can comment about this.

I don't have an idea right away how to fix this correctly. Probably it would require quite a big chance into the audit logging system, where we would have to somehow keep track of all the services that were "current" during the GraphQL field resolution, and then somehow magically pick the correct current service when the Promises get settled and audit log events are produced. I don't know how that could be implemented.

I'm suggesting that for now we rely on users of these APIs not acting on behalf of multiple services during a GraphQL request and that would produce correct "actor services" in the audit log. Let's get rid of the `serviceType` arguments ASAP and then we won't have this issue any more.

akikoskinen: > Darn, this actually revealed another problem with the audit log implementation.

...or perhaps it's a problem with the GraphQL API with the `serviceType` arguments. Anyway, Houston, we have a problem.

mikkokeskinen: Ah. Maybe leave it as is and we'll re-implement the logging properly soon?

mikkokeskinen (approved):
👍

PR #286, opened 2021-03-09
Don't report errors when verified personal information (or parts of it) are missing

akikoskinen:
Missing data is considered a normal situation.


mikkokeskinen (approved):
LGTM

PR #287, opened 2021-03-10
Change audit log date_time_epoch to ms precision

vikoivun:


profiles/tests/test_audit_log.py
```diff
@@ -50,7 +50,9 @@ def assert_common_fields(
     assert audit_event["operation"] == operation
     assert audit_event["actor"]["role"] == actor_role
 
-    assert_almost_equal(audit_event["date_time_epoch"], now)
+    # 2000ms limit was arrived to by checking the maximum
+    # time at local laptop (~1000ms)
+    assert_almost_equal(audit_event["date_time_epoch"], now * 1000, 2000)
```
akikoskinen, line 55: These tests shouldn't use the `get_unix_timestamp_now` function anymore, because it returns only  a second precision timestamp. Instead something more accurate reference time should be used as `now`.

akikoskinen (changes requested):

profiles/tests/test_audit_log.py
vikoivun: > These tests shouldn't use the `get_unix_timestamp_now` function anymore, because it returns only a second precision timestamp. Instead something more accurate reference time should be used as `now`.

Good point. Such coarse testing would have hidden missing ms-component entirely. How about this, just switching to datetime?

akikoskinen: Yep.

profiles/tests/test_audit_log.py
```diff
@@ -42,17 +41,19 @@ def assert_common_fields(
...
     audit_event = log_message["audit_event"]
 
     assert audit_event["origin"] == "PROFILE-BE"
     assert audit_event["status"] == "SUCCESS"
     assert audit_event["operation"] == operation
     assert audit_event["actor"]["role"] == actor_role
 
-    assert_almost_equal(audit_event["date_time_epoch"], now)
+    # 50ms limit was arrived to by checking the maximum time at local laptop
+    assert_almost_equal(audit_event["date_time_epoch"], now_ms_timestamp, 50)
```
akikoskinen, line 55: The comment is quite obvious. These kind of leeways usually come from the hat more or less. No need to state it out loud.

But you could put the magic number 50 into a well named constant. And perhaps also use the same on the line 60 in the other timestamp check.

akikoskinen (approved):
Yeah. Squash the commits and remember to prefix the commit message with the ticket number.

akikoskinen (approved):
Super 👍

PR #288, opened 2021-03-10
Make primary email optional

mikkokeskinen:
Make primary email optional again in the admin, schema and the import. I guess in the future there could be a way to make it mandatory by service, as some existing services depend on this.

Refs HP-476


profiles/tests/test_models.py
```diff
@@ -173,27 +175,8 @@ def test_import_customer_data_with_missing_email():
...
-def test_validation_should_fail_with_no_primary_email(profile):
-    with pytest.raises(ProfileMustHaveOnePrimaryEmail):
-        validate_primary_email(profile)
-
-
-def test_validation_should_fail_with_multiple_primary_emails(profile):
-    EmailFactory(profile=profile, primary=True)
-    EmailFactory(profile=profile, primary=True)
-    with pytest.raises(ProfileMustHaveOnePrimaryEmail):
-        validate_primary_email(profile)
```
akikoskinen, line 196: I suppose we don't want to allow more than one primary email still.

profiles/tests/test_gql_update_my_profile_mutation.py
```diff
@@ -673,6 +708,56 @@ def test_normal_user_can_remove_email(rf, user_gql_client, email_data):
...
+    )
+
+    expected_data = {"updateMyProfile": {"profile": {"emails": {"edges": []}}}}
+
+    mutation = t.substitute(
+        primary_email_id=to_global_id(type="EmailNode", id=primary_email.id),
+        email_id=to_global_id(type="EmailNode", id=email.id),
+        email=email_data["email"],
+        email_type=email_data["email_type"],
+        primary=str(email_data["primary"]).lower(),
```
akikoskinen, line 755: The template doesn't contain these three placeholders. After removing these, the `email_data` fixture isn't needed in this test anymore.

profiles/tests/test_gql_update_my_profile_mutation.py
```diff
@@ -93,6 +93,41 @@ def test_normal_user_can_update_profile(rf, user_gql_client, email_data, profile
...
+                updateMyProfile(
+                    input: {
+                        profile: {
+                            nickname: "${nickname}"
+                        }
+                    }
+                ) {
+                    profile{
+                        nickname,
+                    }
```
akikoskinen, line 116: This test could get emails in the response and check that it is (still) empty.

profiles/tests/test_gql_create_my_profile_mutation.py
```diff
@@ -68,7 +66,7 @@ def test_normal_user_can_create_profile(rf, user_gql_client, email_data, profile
     assert dict(executed["data"]) == expected_data
 
 
-def test_normal_user_cannot_create_profile_with_no_primary_email(
+def test_normal_user_can_create_profile_with_no_primary_email(
```
akikoskinen, line 69: This test is now identical to the previous one, except that in the first one the email is primary, and in this one it's not. These tests could be merged into one parameterized test function.

akikoskinen (changes requested):
The main point is that more than one primary email shouldn't probably be accepted even in the future. Otherwise, some smaller comments.

profiles/tests/test_models.py
mikkokeskinen: It was ensured in the admin, GQL, and the importer. But I now added the check to the Profile model too.

akikoskinen: Although I commented on some place in the code, I intention was that it should be ensured that by no means it's possible to create multiple primary emails for a profile. Perhaps my main concern was the GraphQL API, although the test code I inserted the comment on isn't directly about the GraphQL API. This was badly communicated by me.

But what led to the comment was that it turns out that I wasn't quite aware of how the GraphQL works when multiple emails are set as primary. For example the GraphQL query:

```graphql
mutation {
  updateMyProfile(input: {
    profile: {
      addEmails: [
        {
          email: "new_email@example.com",
          emailType: OTHER
          primary: true
        }
      ]
      updateEmails: [
        {
          id: "AAAA",
          primary: true
        },
        {
          id: "BBBB",
          primary: true
        },
      ]
    }
  }) {
    profile {
      ...
    }
  }
}
```

I thought that this would result into some error, telling the user not try to set multiple emails as primary. But it seems that instead, the query (was and still) is accepted, and whatever email happens to be the last one processed, wins and will be the primary email. A bit weird design, but it is what it is.

So, based on that, my original concern wasn't that valid after all.

akikoskinen: In other words, as the `PROFILE_MUST_HAVE_ONE_PRIMARY_EMAIL` error code was removed in this PR, I was expecting that there should now be some `PROFILE_CAN_NOT_HAVE_MORE_THAN_ONE_PRIMARY_EMAILS` error code. But now I have learned, that there doesn't have to be.

akikoskinen (approved):
Looks good otherwise, but the GraphQL snippets in tests could use some indentation fixes.

mikkokeskinen:
This one?

```
    emails{
        edges{
        node{
            email,
            emailType,
            primary,
            verified
        }
        }
```

I thought it was intentional because the existing test above used that indentation.

profiles/tests/test_gql_create_my_profile_mutation.py
```diff
@@ -62,46 +65,40 @@ def test_normal_user_can_create_profile(rf, user_gql_client, email_data, profile
...
                 profile{
-                    id
+                    emails{
+                        edges{
+                        node{
+                            email,
+                            emailType,
+                            primary,
+                            verified
+                        }
```
akikoskinen, line 93: Indentation.

profiles/tests/test_gql_update_my_profile_mutation.py
```diff
@@ -93,6 +93,53 @@ def test_normal_user_can_update_profile(rf, user_gql_client, email_data, profile
...
+                    emails{
+                        edges{
+                        node{
+                            email
+                            emailType
+                            primary
+                            verified
+                        }
+                        }
+                    }
```
akikoskinen, line 125: Indentation.

profiles/tests/test_gql_update_my_profile_mutation.py
```diff
@@ -673,6 +720,53 @@ def test_normal_user_can_remove_email(rf, user_gql_client, email_data):
...
+                        node{
+                            id
+                            email
+                            emailType
+                            primary
+                        }
+                        }
+                    }
+                }
+            }
```
akikoskinen, line 755: Indentation.

akikoskinen:
> I thought it was intentional because the existing test above used that indentation.

Ok. I suppose it's not intentional. Although it's off course purely a stylish thing (unlike in the Python code). But perhaps it improves readability when indentation flows fluently in the GraphQL snippets too.

akikoskinen (approved):
Nice

PR #289, opened 2021-03-11
Simplify graphql testing

akikoskinen:
Move some GraphQL request test setup into the test client. Namely it now constructs an HTTP request object (used as the `context` in the GraphQL query handling), and inserts a `User`, a `UserAuthorization`, and a `Service` into it. This mimics more closely the way how requests are handled in the real world for **all** tests now.

There are a few quite big commits in this PR, but they mostly just repeat the same 3-4 line change (mostly removing lines) over and over again throughout the test code.

### DIFF
--- a/open_city_profile/tests/conftest.py
+++ b/open_city_profile/tests/conftest.py
@@ -7,6 +7,7 @@ from django.core.management import call_command
 from django.test import RequestFactory
 from graphene.test import Client as GrapheneClient
 from graphql import build_client_schema, introspection_query
+from helusers.authz import UserAuthorization
 
 from open_city_profile.graphene import GQLDataLoaders
 from open_city_profile.schema import schema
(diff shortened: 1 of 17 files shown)


open_city_profile/tests/conftest.py
```diff
@@ -18,13 +19,21 @@
...
         """
-        return super().execute(*args, middleware=[GQLDataLoaders()], **kwargs)
+        if context is None:
+            context = RequestFactory().post("/graphql")
+
+        if not hasattr(context, "user") and hasattr(self, "user"):
+            context.user = self.user
+
+        return super().execute(
+            *args, context=context, middleware=[GQLDataLoaders()], **kwargs
```
mikkokeskinen, line 50: I know this execute call wasn't added in this PR, but should this be changed to something like
```python
kwargs.setdefault("middleware", [GQLDataLoaders()])
return super().execute(
    *args, context=context, **kwargs
)
```

So that the middleware can be customised if needed?

akikoskinen: Hmm, good question. If this was a "general use client", then perhaps it would be appropriate. As it's used only for a specific need in the tests, I don't think it makes much difference. I think it can be changed later when/if needed.

mikkokeskinen (approved):
LGTM

PR #291, opened 2021-03-12
View some relations as autocomplete fields

mikkokeskinen:
Django Admin automatically populates a dropdown for relations, but when there are thousands of instances the dropdown generation generates many SQL queries.

By displaying the fields as autocomplete the dropdown is not generated right away.

Refs HP-561

### DIFF
--- a/profiles/admin.py
+++ b/profiles/admin.py
@@ -43,6 +43,7 @@ class RepresentativeAdmin(admin.StackedInline):
     fk_name = "representative"
     verbose_name = "Representative"
     verbose_name_plural = "Representing"
+    autocomplete_fields = ("representee",)
 
 
 class RepresenteeAdmin(admin.StackedInline):
@@ -51,6 +52,7 @@ class RepresenteeAdmin(admin.StackedInline):
     fk_name = "representee"
     verbose_name = "Representative"
     verbose_name_plural = "Represented by"
+    autocomplete_fields = ("representative",)
 
 
 class AlwaysChangedModelForm(ModelForm):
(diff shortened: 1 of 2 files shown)


profiles/admin.py
```diff
@@ -177,6 +179,7 @@ class ExtendedProfileAdmin(VersionAdmin):
     ]
     change_list_template = "admin/profiles/profiles_changelist.html"
     list_filter = ("service_connections__service",)
+    raw_id_fields = ("user",)
```
charn, line 182: What about `autocomplete_fields` ? :thinking:

mikkokeskinen: Louhis you are a genius! Why didn't I know of that. 😄

profiles/tests/test_admin.py
```diff
@@ -54,3 +56,19 @@ def test_profile_should_not_be_valid_with_two_or_more_primary_emails(profile):
...
+    admin_client, django_assert_max_num_queries, profile
+):
+    view_profile_url = reverse("admin:profiles_profile_change", args=(profile.pk,))
+
+    with django_assert_max_num_queries(50):
+        admin_client.get(view_profile_url)
+
+    # Add 100 profiles
+    for i in range(0, 100):
+        ProfileFactory()
```
akikoskinen, line 71:
```suggestion
    ProfileFactory.create_batch(100)
```

akikoskinen (commented):
Nice, with a test and everything.

akikoskinen (approved):

charn (approved):
LGTM! :+1: :shipit:

PR #295, opened 2021-03-15
Verified personal information to all graphql queries

akikoskinen:
The verified personal information is now present in the main `Profile` GraphQL type, `ProfileNode`. The `myProfile` query specific return type `ProfileWithVerifiedPersonalInformationNode` was removed and `myProfile` once again returns a `ProfileNode`.

When the requester is not querying their own `Profile`, they need a `services.can_view_verified_personal_information` permission.

When the requester **is** querying their own `Profile`, the "loa" claim of the authentication JWT is checked as before. But the "loa" claim check doesn't apply to the "querying others' profiles" case.

The `profiles` query's search criteria is not expanded in this PR as specified in HP-550. I'll create a separate PR for that.


profiles/tests/test_gql_profile_query.py
```diff
@@ -276,7 +278,10 @@ def test_staff_user_needs_required_permission_to_access_verified_personal_inform
         )
     )
 
-    executed = user_gql_client.execute(query, service=service)
+    token_payload = {"loa": "high" if has_loa else "low"}
```
mikkokeskinen, line 281: We could maybe test with "substantial" (or both), because I guess the loa won't be "high" in majority of the cases.

akikoskinen: Yep. Although this isn't the main "loa" test, it's just here for checking that the "loa" doesn't let you access others' profiles.

akikoskinen: Actually parametrizing the test by the "has_loa" argument is useless. The "loa" claim can always be included and it will cover all cases worth testing (in this test).

mikkokeskinen: True. Didn't notice that this was about the staff user.

akikoskinen: Changed the "loa" to "substantantial" and removed the parametrization, i.e. the "loa" claim is always present in this particular test.

mikkokeskinen (approved):
LGTM 👍

PR #299, opened 2021-03-19
Audit log actor service's name

akikoskinen:
The obsolete `service_type` field of `Service` is not used for audit logging anymore. Instead the `name` is used, which is a mandatory field in `Service`.

Also the actor's service information is now included within the "actor" element in the audit log. There is no "actor_service" element anymore.


mikkokeskinen (approved):
👍

PR #300, opened 2021-03-19
More audit log changes

akikoskinen:
- `user_name` removed from `actor` and `target` elements.
- moved `ip_address` into the `actor` element  (was previously under the `profilebe` element).
- in the `target` element:
  - `profile_id` renamed to `id`
  - `profile_part` renamed to `type`
- the `profilebe` element doesn’t exist anymore.


mikkokeskinen (approved):
👍

PR #303, opened 2021-03-22
Show version in admin index page footer

akikoskinen:
This makes it easier to see what version of the software is running.

### DIFF
--- a/open_city_profile/admin_site.py
+++ b/open_city_profile/admin_site.py
@@ -1,0 +1,13 @@
+from django.conf import settings
+from helusers.admin_site import AdminSite as HelUsersAdminSite
+
+
+class AdminSite(HelUsersAdminSite):
+    index_template = "admin/admin_index.html"
+
+    def each_context(self, request):
+        context = super().each_context(request)
+        context["version"] = (
+            settings.VERSION if settings.VERSION is not None else "unknown"
+        )
+        return context
--- a/templates/admin/admin_index.html
+++ b/templates/admin/admin_index.html
@@ -1,0 +1,6 @@
+{% extends "admin/index.html" %}
+{% load i18n %}
+
+{% block footer %}
+<div id="footer">{% trans "Version" %}: {{ version }}</div>
+{% endblock %}
--- a/open_city_profile/apps.py
+++ b/open_city_profile/apps.py
@@ -1,6 +1,11 @@
 from django.apps import AppConfig
+from django.contrib.admin.apps import AdminConfig
 
 
 class OpenCityProfileConfig(AppConfig):
     name = "open_city_profile"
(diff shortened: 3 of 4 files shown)


open_city_profile/admin_site.py
```diff
@@ -0,0 +1,11 @@
...
+from helusers.admin_site import AdminSite as HelUsersAdminSite
+
+
+class AdminSite(HelUsersAdminSite):
+    index_template = "admin/admin_index.html"
+
+    def each_context(self, request):
+        ret = super().each_context(request)
+        ret["version"] = settings.VERSION or "unknown"
+        return ret
```
mikkokeskinen, line 11: This kind of variable assign is a bit of a javascriptism. Even in JS it's not advisable. Also the return variable name could be `context`.

```suggestion
    def each_context(self, request):
        context = super().each_context(request)
        context["version"] = (
            settings.VERSION if settings.VERSION is not None else "unknown"
        )
        return context
```

templates/admin/admin_index.html
```diff
@@ -0,0 +1,6 @@
+{% extends "admin/index.html" %}
+{% load i18n %}
+
+{% block footer %}
+<div id="footer">{% trans "Version" %}: {{ version }}</div>
```
mikkokeskinen, line 5: Would it be clearer to label this "git hash" or something like that? If the VERSION is changed to proper version number some day it could be then changed back?

akikoskinen: This project doesn't have an opinion about what the user of this software might put into the `VERSION` environment variable, where `settings.VERSION` is read from. Although `settings.VERSION` defaults to the output of a `git describe` command (but that's not a git hash either). So in that sense I would keep this as it is.

Or then make the whole string configurable, so one can input anything they want into it. But that sounds quite an overkill.

mikkokeskinen: I'm just saying that it would be nice to know exactly what that is. As it will be the result of git describe, could it say that it's a result of `git describe` somewhere? But maybe it's just something our team should know or maybe it should be documented in Confluence.

open_city_profile/admin_site.py
akikoskinen: Changed.

templates/admin/admin_index.html
akikoskinen: The `git describe` is a backup plan (and IMO could even be removed from the code). The main way to use this is to set the `VERSION` environment variable. But within this code base we can't control what value anybody decides to put in that environment variable.

mikkokeskinen (approved):
👍

charn (approved):
LGTM! :+1: :shipit:

PR #304, opened 2021-03-24
Add audit logging to verified personal information parts

akikoskinen:
When `Profile`'s `VerifiedPersonalInformation`, or any of its addresses, is touched (in CRUD terms), a line is added to the audit log.

The production code changes are quite small. The tests however experienced a bigger refactoring.

### DIFF
--- a/profiles/audit_log.py
+++ b/profiles/audit_log.py
@@ -49,7 +49,11 @@ def log(action, instance):
 
         current_time = datetime.now(tz=timezone.utc)
         current_user = get_current_user()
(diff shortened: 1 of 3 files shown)


mikkokeskinen (approved):
Looks good!

Especially making the audit log test fixture more clear. It took me a good while to decipher earlier.

akikoskinen:
Sonar warned about two places where variables were assigned but never read. And it was right. So I removed those useless lines. That's why the second to last commit was updated.

PR #307, opened 2021-03-29
Avoid unnecessary database updates with empty addresses

akikoskinen:
The previous version of the code created or updated empty address models of `VerifiedPersonalInformation` into the database unnecessarily. Not any more.

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -899,24 +899,23 @@ class CreateOrUpdateProfileWithVerifiedPersonalInformationMutation(graphene.Muta
 
         for address_type in address_types:
             address_input = address_type["input"]
-            if address_input:
-                address_name = address_type["name"]
-                address_model = address_type["model"]
-                try:
-                    address = getattr(information, address_name)
-                    for field, value in address_input.items():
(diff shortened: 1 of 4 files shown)


profiles/schema.py
```diff
@@ -904,9 +904,13 @@ def mutate(parent, info, input):
                 address_model = address_type["model"]
```
mikkokeskinen, line 904: Good change. Although I would do the address thing something like this:

```python
        for address_type in address_types:
            address_input = address_type["input"]
            if not address_input:
                continue

            address_name = address_type["name"]
            address_model = address_type["model"]
            try:
                address = getattr(information, address_name)
            except address_model.DoesNotExist:
                address = address_model(verified_personal_information=information)
                
            for field, value in address_input.items():
                setattr(address, field, value)

            if not address.is_empty():
                address.save()
            else:
                if address.id:
                    address.delete()
```

akikoskinen: That indeed looks more straightforward.

mikkokeskinen (approved):
👍

PR #310, opened 2021-04-06
Add primary email input to createOrUpdateProfileWithVerifiedPersonalInformation mutation

akikoskinen:
It's now possible to create/update the primary email of a profile with the `createOrUpdateProfileWithVerifiedPersonalInformation` mutation.

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -890,12 +890,15 @@ class CreateOrUpdateProfileWithVerifiedPersonalInformationMutation(graphene.Muta
     def _handle_primary_email(profile, primary_email_input):
         email_address = primary_email_input["email"]
 
-        email = profile.emails.create(email=email_address, email_type=EmailType.NONE)
+        email, email_created = profile.emails.get_or_create(
+            email=email_address, defaults={"email_type": EmailType.NONE}
+        )
 
-        profile.emails.update(primary=False)
+        profile.emails.exclude(pk=email.pk).filter(primary=True).update(primary=False)
 
-        email.primary = True
-        email.save()
+        if not email.primary:
+            email.primary = True
+            email.save()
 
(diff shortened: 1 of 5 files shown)


tituomin (approved):

profiles/schema.py
```diff
@@ -874,16 +874,26 @@ class Arguments:
...
+                email = Email(
+                    profile=profile,
+                    email=email_address,
+                    email_type=EmailType.NONE,
+                    primary=True,
+                )
+
+            if email:
+                profile.emails.filter(primary=True).update(primary=False)
+                email.save()
```
mikkokeskinen, line 896: I would try to make this a bit clearer. Maybe like:

```python
email, email_created = profile.emails.get_or_create(email=email_address, defaults={
    "email_type": EmailType.NONE,
})

profile.emails.exclude(pk=email.pk).update(primary=False)

if not email.primary:
    email.primary = True
    email.save()
```

I didn't test this, but something like this should work.

Also I'm not a big fan of always calling this method and then checking if an email was provided. At the very least the check could be

```python
if not primary_email_input:
    return
```

akikoskinen: > Also I'm not a big fan of always calling this method and then checking if an email was provided.

So, would it be better if the checking would take place at the calling site, before this function is called?

mikkokeskinen: I think so yes. Then when reading that function you would know if something will happen to the emails or not. And don't need to jump to this function to check.

akikoskinen: Updated now as suggested, as it was a good one. The `mutate` function was getting too complex (according to flake8) so I inserted the commit a44724b to make it simpler.

mikkokeskinen (approved):

PR #312, opened 2021-04-12
HP-645 Make national_identification_number fields hash_key a callable

mikkokeskinen:
Before the hash_key value was a variable that would get evaluated when the field was declared creating a new migration for that field every time the hash_key changed. (e.g. in different environments)

Now the function name gets added to the migrations and won't change even if the SALT_NATIONAL_IDENTIFICATION_NUMBER setting changes.

Refs HP-645


akikoskinen (approved):
Looks good.

This would be a good candidate for contributing upstream. There's this [one (closed) issue](https://gitlab.com/guywillett/django-searchable-encrypted-fields/-/issues/10) that touches the subject, but it doesn't seem to realize this problem that is fixed in this PR.

PR #313, opened 2021-04-19
Add one more permission to test setup

akikoskinen:
This test uses two Services, service_berth and service_youth. The idea
is that if the requester is coming from service_youth, then they have
all the required permissions to create a Profile with sensitive data. If
the requester is coming from servie_berth, then they can't, since in
that case they don't have the required sensitive data permissions (just
the "can_manage_profiles" permission).

Since the request in this test is executed as coming from service_berth,
the request fails and that is the expected result. If the request would
be coming from service_youth, then it should succeed. But that wasn't
the case, without this one missing permission.


mikkokeskinen (approved):
👍

PR #314, opened 2021-04-19
Deactivate service type GraphQL arguments

akikoskinen:
The `serviceType` arguments in some GraphQL calls were deprecated earlier, but left still functioning. Since “azp” JWT claims can now be used for service determination everywhere, it’s time to make the `serviceType` arguments non-functional.


mikkokeskinen (approved):
Looks good to me!

PR #315, opened 2021-04-20
Add client id to audit log

akikoskinen:
The client id is received from the HTTP request's authentication JWT, from its "azp" claim. That value (if present) is now added to the audit log lines.


mikkokeskinen (approved):
Good stuff

PR #316, opened 2021-04-20
Cleaning up service types part 1 (of 2)

akikoskinen:
The `Service`'s `service_type` field is obsolete. It can be removed from the code base, except the few places where it still remains for backwards compatibility.

The changes in this PR are those that were needed to achieve the last commit, which removes the `service_type = ServiceType.BERTH` from `ServiceFactory`. The changes seem to mostly concern the tools and utilities.

There are many places where `service_type` is still used even after this PR. But those are subject for another PR(s).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,12 +73,10 @@ Prerequisites:
 5. Set permissions for service staff members if needed:
    * Create group(s) (via Django admin) and add user(s) to the group
    * Create service permissions for group manually via Django admin or for example:
-     * `docker exec profile-backend python manage.py add_object_permission BERTH
-     VeneAdmin can_view_profiles`
-     * where:
-       * `service_type=BERTH`
-       * `group_name=VeneAdmin`
-       * `permission=can_view_profiles`
+     * `docker exec profile-backend python manage.py add_object_permission ServiceName GroupName can_view_profiles`,  where:
+       * `ServiceName` is the name of the Service the permission is given for
+       * `GroupName` is the name of the group to whom the permission is give
+       * `can_view_profiles` is the name of the permission
    * Permissions can be removed as follows:
      * `docker exec profile-backend python manage.py remove_object_permission BERTH
      VeneAdmin can_view_profiles`
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Prerequisites:
     * Command will generate:
       * All available services
       * One group per service (with `can_manage_profiles` permissions)
-      * One user per group (with username `{service_type}_user`)
+      * One user per group (with username `{group.name}_user`)
       * Profiles
         * With user
         * With email, phone number and address
--- a/README.md
+++ b/README.md
@@ -78,8 +78,7 @@ Prerequisites:
        * `GroupName` is the name of the group to whom the permission is give
        * `can_view_profiles` is the name of the permission
    * Permissions can be removed as follows:
-     * `docker exec profile-backend python manage.py remove_object_permission BERTH
-     VeneAdmin can_view_profiles`
+     * `docker exec profile-backend python manage.py remove_object_permission ServiceName GroupName can_view_profiles`
 
 6. Seed development data
     * **Note!** This command will flush the database.
--- a/profiles/admin.py
+++ b/profiles/admin.py
@@ -3,7 +3,7 @@ from functools import reduce
 
 from django import forms
 from django.contrib import admin, messages
-from django.core.exceptions import PermissionDenied
+from django.core.exceptions import PermissionDenied, ValidationError
 from django.forms.models import ModelForm
 from django.http import JsonResponse
 from django.shortcuts import render
@@ -24,6 +24,7 @@ from profiles.models import (
     VerifiedPersonalInformation,
 )
 from services.admin import ServiceConnectionInline
+from services.models import Service
 from subscriptions.admin import SubscriptionInline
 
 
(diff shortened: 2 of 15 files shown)


profiles/admin.py
```diff
@@ -197,7 +204,8 @@ def upload_json(self, request):
         try:
             if request.method == "POST":
                 data = json.loads(request.FILES["json_file"].read())
-                result = Profile.import_customer_data(data)
+                service_name = request.POST["service_name"]
```
mikkokeskinen, line 207: Now that there is a non-file field in the form, the form handling could be done like it's usually done with Django forms.

i.e. instantiating the form, calling is_valid and using form.cleaned_data.

profiles/models.py
```diff
@@ -180,6 +179,7 @@ def import_customer_data(cls, data):
         And returns dict where key is the customer_id and value is the UUID of created profile object
         """
         result = {}
+        service = Service.objects.filter(name=service_name).first()
```
mikkokeskinen, line 182: I would prefer checking if the service_name is given and explicitly getting the one.

utils/utils.py
```diff
@@ -128,25 +141,18 @@ def generate_services():
     Also assigns allowed data fields for each created service.
     """
     services = []
-    for service_type in ServiceType:
-        service = Service.objects.filter(service_type=service_type).first()
+    for service_spec in SERVICES:
+        service_name = service_spec["name"]
+        service = Service.objects.filter(name=service_name).first()
```
mikkokeskinen, line 152: Same filter() vs get() here. In my opinion when getting one explicit thing it's more idiomatic Django to use
```python
try:
    instance = Model.objects.get(key=value)
except Model.objects.DoesNotExist:
    instance = Model(...)
```

And maybe handle MultipleObjectsReturned exception where appropriate.

README.md
```diff
@@ -73,11 +73,10 @@ Prerequisites:
 5. Set permissions for service staff members if needed:
    * Create group(s) (via Django admin) and add user(s) to the group
    * Create service permissions for group manually via Django admin or for example:
-     * `docker exec profile-backend python manage.py add_object_permission BERTH
-     VeneAdmin can_view_profiles`
+     * `docker exec profile-backend python manage.py add_object_permission ServiceName GroupName can_view_profiles`
      * where:
-       * `service_type=BERTH`
-       * `group_name=VeneAdmin`
+       * `service=ServiceName`
```
mikkokeskinen, line 78: This was a strange help text to begin with, but maybe switch the order?
```
ServiceName = The service
GroupName = The name of the group you want to give permissions to
can_view_profiles = The permission
``` 
Or something like that.

services/management/commands/object_permission_command.py
```diff
@@ -0,0 +1,43 @@
...
+            if kwargs["permission"] not in available_permissions:
+                raise ValueError
+            service_name = kwargs["service"]
+            service = Service.objects.get(name=service_name)
+            group = Group.objects.get(name=kwargs["group_name"])
+            permission = kwargs["permission"]
+
+            self.do_handle(permission, group, service)
+        except Service.DoesNotExist:
+            self.stdout.write(self.style.ERROR("Invalid service given"))
```
mikkokeskinen, line 39: Could maybe use [CommandError](https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#django.core.management.CommandError) here. I think tests can then look for that too.

mikkokeskinen (changes requested):
Great to get rid of the service type. Gave some minor comments.

profiles/models.py
akikoskinen: Changed

utils/utils.py
akikoskinen: Changed

README.md
akikoskinen: Improved

services/management/commands/object_permission_command.py
akikoskinen: Changed

profiles/admin.py
akikoskinen: Sounds like a good direction. I'm not that familiar with Django forms, so have to study a bit what is it exactly that should be done :monocle_face:

mikkokeskinen: [Here](https://docs.djangoproject.com/en/2.2/topics/forms/#the-view) is an example in the Django docs.

akikoskinen: Actually, I started thinking, that my time is probably better spent doing something else. While this would be a good change, perhaps it's not a good time to do it now. I don't even know whether this form has any use (any more). So I'm suggesting not to change this right now.

mikkokeskinen: It shouldn't take too long. Something like this:

```python
    @method_decorator(superuser_required, name="dispatch")
    def upload_json(self, request):
        try:
            if request.method == "POST":
                form = ImportProfilesFromJsonForm(request.POST, request.FILES)
                if form.is_valid():
                    data = json.loads(request.FILES["json_file"].read())
                    service_name = form.cleaned_data["service_name"]
                    result = Profile.import_customer_data(data, service_name)
                    response = JsonResponse(result)
                    response["Content-Disposition"] = "attachment; filename=export.json"

                    return response
                else:
                    raise ValidationError(', '.join(form.errors.values()))
            else:
                form = ImportProfilesFromJsonForm()
                return render(request, "admin/profiles/upload_json.html", {
                    "form": form
                })
        except Exception as err:
            messages.error(request, err)
            form = ImportProfilesFromJsonForm()
            return render(request, "admin/profiles/upload_json.html", {
                "form": form
            })
```

akikoskinen: Allrighty, not that big a change, so bit the bullet. The form conveniently gives the actual `Service` directly, so no need to fetch it explicitly.

mikkokeskinen (approved):
Good job!

PR #319, opened 2021-04-30
WIP: prototyping input validation

akikoskinen:
Here's the current state with prototyping the GraphQL input validation. This uses [a fork](https://github.com/City-of-Helsinki/graphene-validator/tree/wip-prototyping) of the [graphe-validator](https://github.com/chpmrc/graphene-validator) library.

The four first commits are actually "production quality" and could be merged as such. The actual prototype is in the fifth commit.

The current prototype uses field level validations. Meaning it checks each input field separately. But just before pushing this WIP branch, I figured that perhaps this system could use the Django model's `full_clean` function, which checks the individual fields, and more. I'll keep on prototyping and see if that solution would be leaner.


akikoskinen:
> The current prototype uses field level validations. Meaning it checks each input field separately. But just before pushing this WIP branch, I figured that perhaps this system could use the Django model's `full_clean` function, which checks the individual fields, and more. I'll keep on prototyping and see if that solution would be leaner.

The model level `full_clean` approach didn't bring instant gratification. One big downside that came apparent quickly is that when using the object type class level `validate` function, `graphene-validator`  can't fill in the `path` data (which tells the path to the input field that failed validation) in to the error.

akikoskinen:
This PR is now superseded by #329.

PR #322, opened 2021-05-04
Validate verified personal information free text fields' allowed character set

akikoskinen:
Free text fields are those that don't have any structural or syntactical limitations. They have a maximum length (not relevant in this PR) and allowed character set (the point of this PR).

The allowed character set for these fields is [Basic Latin (block)](https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)) + [Latin-1 Supplement (block)](https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)) - [Control Code (general category=Cc)](https://en.wikipedia.org/wiki/C0_and_C1_control_codes). The fields' values are automatically [normalized to the "Normalization Form Canonical Composition" form](https://en.wikipedia.org/wiki/Unicode_equivalence#Normal_forms) before validation. This process combines any character sequences with a combining character into the combined form. The allowed character set contains many combined characters but not the decomposed forms.

The fields are validated using the Django model field validation machinery. That works directly in the admin UI (although none of these fields are editable there). The validators also kick in e.g. when the model instance's `full_clean` method gets called.


akikoskinen:
Adding validators to model fields makes Django want to create a migration file. The migration file wouldn't make any changes to the database though. There isn't any migration file included in this PR. There will be more changes to the relevant models' fields in other PRs as well. Django will then again create migration files, and at least some of those will be similar no-op migrations.

@mikkokeskinen @tituomin What do you think, what should be done with these kind of migration files? Should all of them be included in the related commits? Just add one final migration file in the last commit that changes the models/fields? Somehow gather all model/field changes into a single commit that could then also include the one related migration file? Other options?

mikkokeskinen:
In my opinion this is a bit too complex. I wouldn't change the fields and wouldn't create unicode helper functions.

I would just add a validator something like this and add it to the fields:

```python
def validate_visible_latin_characters_only(value: str) -> None:
    if not re.fullmatch('[\u0020-\u00ff]*', unicodedata.normalize("NFC", value)):
        raise Exception("String contains unaccepted characters")
```

akikoskinen:
> In my opinion this is a bit too complex. I wouldn't change the fields and wouldn't create unicode helper functions.
> 
> I would just add a validator something like this and add it to the fields:
> 
> ```python
> def validate_visible_latin_characters_only(value: str) -> None:
>     if not re.fullmatch('[\u0020-\u00ff]*', unicodedata.normalize("NFC", value)):
>         raise Exception("String contains unaccepted characters")
> ```

I also thought (after I had written all these helpers) the validator could be simplified to one simple function. I do kind of like the helper functions though as their names give meanings to otherwise obscure code point ranges (this can off course also be documented for example as comments, but I generally like meaningful names over comments).

What comes to the normalization: the idea was that the NFC normalized form _also goes to the database_. It can off course be discussed should this actually be done like that. The implementation was chosen to do that. Doing the normalization only within a validator wouldn't affect the value stored into the database.

mikkokeskinen:
Our documentation only mentions normalization when validating. Maybe there has been a separate decision about saving? Although I would save the string as-is. What do you think @tituomin?

Nevertheless, I'm not sure if `to_python` is the correct method? Maybe it should be done in `get_db_prep_value` or even `pre_save`?

mikkokeskinen:
About giving names to things, isn't the method name (`validate_visible_latin_characters_only`) already an apt name?

akikoskinen:
> Our documentation only mentions normalization when validating. Maybe there has been a separate decision about saving? Although I would save the string as-is. What do you think @tituomin?

I suppose there hasn't been any discussion about this. I just thought it would be a good idea to have such data in the database that as such passes the validations for that field. There are off course other options that don't necessarily need any normalization. For example also allow the [Combining Diacritical Marks block](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks). Or just deny all strings that have combining characters and accept only NFC normalized input.

> Nevertheless, I'm not sure if `to_python` is the correct method? Maybe it should be done in `get_db_prep_value` or even `pre_save`?

Both `get_db_prep_value` and `pre_save` are only relevant when saving. Validation can take place also without saving (and in the upcoming GraphQL input validation that's indeed the use case). If normalization _for database_ happens in the saving phase then validators would have to handle "unnormalized" strings in some way. One option to do that handling would be to normalize the string. But then every validator (that cares about this) would have to do that separately. And in this case the normalization would be done twice for same data, first in the validator and then again in the saving phase. I used `to_python` in order to avoid that.

akikoskinen:
> About giving names to things, isn't the method name (`validate_visible_latin_characters_only`) already an apt name?

Indeed it is quite explaining :smile:

mikkokeskinen:
> Both get_db_prep_value and pre_save are only relevant when saving.

I thought that was exactly the point. Save the string normalized.

> Validation can take place also without saving (and in the upcoming GraphQL input validation that's indeed the use case). If normalization for database happens in the saving phase then validators would have to handle "unnormalized" strings in some way.

I don't quite understand this. Could you elaborate?

> And in this case the normalization would be done twice for same data, first in the validator and then again in the saving phase. I used to_python in order to avoid that.

I'm not following this either. Surely validation shouldn't change the data in any way?

akikoskinen:
> > Both get_db_prep_value and pre_save are only relevant when saving.
> 
> I thought that was exactly the point. Save the string normalized.

Yes, that's one point. Another point is to avoid normalizing the same data more than once, for performance reasons. So if normalization _has to_ be also performed in the validation phase, then doing it again in the saving phase is wasteful.

> > Validation can take place also without saving (and in the upcoming GraphQL input validation that's indeed the use case). If normalization for database happens in the saving phase then validators would have to handle "unnormalized" strings in some way.
> 
> I don't quite understand this. Could you elaborate?

Actually the GraphQL case isn't special. It will be somewhat analogous to how Django forms work so we can think of those. When Django form gets some input data from the user, it performs validation, maybe with a call to the form's `is_valid()` method. If validation reports errors, those are reported to the user and no saving happens. That's what I meant with "validation can take place also without saving". Perhaps it wasn't very clearly put.

> > And in this case the normalization would be done twice for same data, first in the validator and then again in the saving phase. I used to_python in order to avoid that.
> 
> I'm not following this either. Surely validation shouldn't change the data in any way?

Perhaps we need to be careful with the terms so we don't confuse each other. 

Django documentation talks about "cleaning data" and "validation" (["_Form validation happens when the data is cleaned._"](https://docs.djangoproject.com/en/3.1/ref/forms/validation/)). Both terms are used in the documentation in a manner that I can't say for sure what is "cleaning" and what is "validation" there. My interpretation would be that "cleaning" can change the data, but "validation" can't.

Anyway `to_python()` can definitely change the data. Perhaps it's mostly meant to change the _type_ of the data (for example turn a string into an integer in an `IntField`). But can it do other kinds of changes too besides just type transformation?

`to_python()` is called from `Field`'s `clean()` method, which is supposedly meant to do "cleaning". But what does "cleaning" mean? Is Unicode normalization cleaning? What if the `clean()` method would be overridden instead of `to_python()` and do the normalization there?

mikkokeskinen:
> Yes, that's one point. Another point is to avoid normalizing the same data more than once, for performance reasons. So if normalization _has to_ be also performed in the validation phase, then doing it again in the saving phase is wasteful.

I think it's a case of premature optimization in these kind of "cold spot" low cpu impact situations. But that's not the only reason to do the normalization twice. I think the validation shouldn't modify the data at all, but to validate we have to normalize the value. e.g. what if you use the validator somewhere else and the value is not normalized before calling the validator?

> `to_python()` is called from `Field`'s `clean()` method, which is supposedly meant to do "cleaning". But what does "cleaning" mean? Is Unicode normalization cleaning? What if the `clean()` method would be overridden instead of `to_python()` and do the normalization there?

Yes, I guess that could be the best way. But in a way where the validation is done first, and only after that the value would be normalized (again). i.e.

```python
    # validation
    def validate_visible_latin_characters_only(value: str) -> None:
        if not re.fullmatch('[\u0020-\u00ff]+', unicodedata.normalize("NFC", value)):
            raise Exception("String contains unaccepted characters")

    # cleaning
    def clean(self, value, model_instance):
        value = super().clean(value, model_instance)
        return unicodedata.normalize("NFC", value)
```

All in all I'm still in the position that we shouldn't normalize the value at all outside the validation.

akikoskinen:
> > About giving names to things, isn't the method name (`validate_visible_latin_characters_only`) already an apt name?
> 
> Indeed it is quite explaining smile

Looking at this more closely, especially what the body of this function could contain. The short regex version (I changed the correct regex here):

```python
def validate_visible_latin_characters_only(value: str) -> None:
    if not re.fullmatch("[\u0020-\u007e\u00a0-\u00ff]*", unicodedata.normalize("NFC", value)):
        raise ValidationError("String contains unaccepted characters")
```

The current version:

```python
def validate_visible_latin_characters_only(value: str) -> None:
    def is_valid(char):
        return (
            is_in_basic_latin_block(char) or is_in_latin_1_supplement_block(char)
        ) and not is_control_code(char)

    if not all([is_valid(c) for c in value]):
        raise ValidationError(
            _("String contains unaccepted characters"), code="unaccepted_characters"
        )
```

Which one better tells what "visible latin characters" exactly means here?

tituomin:
> Adding validators to model fields makes Django want to create a migration file. The migration file wouldn't make any changes to the database though. There isn't any migration file included in this PR. There will be more changes to the relevant models' fields in other PRs for this story as well. Django will then again want to create migration files, and at least some of those will be similar no-op migrations.
> 
> @mikkokeskinen @tituomin What do you think, what should be done with these kind of migration files? Should all of them be included in the related commits, resulting in multiple kind-of-needless no-op migration files? Just add one final migration file in the last commit that changes the models/fields? Somehow gather all model/field changes into a single commit that could then also include the one related migration file? Other options?

I say don't worry about the migrations for now, maybe squash them later if there are huge amounts in the end https://docs.djangoproject.com/en/3.2/topics/migrations/#squashing-migrations

> You are encouraged to make migrations freely and not worry about how many you have; the migration code is optimized to deal with hundreds at a time without much slowdown. However, eventually you will want to move back from having several hundred migrations to just a few, and that’s where squashing comes in.

tituomin:
I also think the custom Field and the _Block abstraction are a bit too much.

I prefer this version:
```def validate_visible_latin_characters_only(value: str) -> None:
    if not re.fullmatch("[\u0020-\u007e\u00a0-\u00ff]*", unicodedata.normalize("NFC", value)):
        raise ValidationError("String contains unaccepted characters")
        ```
But it could maybe contain a simple comment naming the blocks present
in the regexp (for those who don't remember them by heart).

I also think the validation functions should probably assume that there hasn't been any processing (normalization) done to the incoming data, so that it's always safe to use it directly with any incoming data in any context.

Whether we should normalize when saving: it could be a separate question with a separate discussion. The most important thing would be that everything in the database passes validation, but that would be the case if validation always normalizes and saving doesn't.

tituomin (commented):
See my comments

akikoskinen:
> > Adding validators to model fields makes Django want to create a migration file. The migration file wouldn't make any changes to the database though. There isn't any migration file included in this PR. There will be more changes to the relevant models' fields in other PRs for this story as well. Django will then again want to create migration files, and at least some of those will be similar no-op migrations.
> > @mikkokeskinen @tituomin What do you think, what should be done with these kind of migration files? Should all of them be included in the related commits, resulting in multiple kind-of-needless no-op migration files? Just add one final migration file in the last commit that changes the models/fields? Somehow gather all model/field changes into a single commit that could then also include the one related migration file? Other options?
> 
> I say don't worry about the migrations for now, maybe squash them later if there are huge amounts in the end https://docs.djangoproject.com/en/3.2/topics/migrations/#squashing-migrations

The scope of the validation changes got more tightly focused to just include validation (as it probably should have been from the beginning, but kind of went on a tangential) and that probably reduces the number of needed migration files. I'm not that worried about the number of migration files any more so my opinion is now that each commit that would produce a migration file, also includes that migration file. That way the commits remain "neat" and complete and can thus more easily be reordered, reverted, cherry-picked if needed and all the other niceties that result from good commits.

mikkokeskinen (approved):
Excellent!

PR #326, opened 2021-05-11
Validate the rest of the verified personal information fields

akikoskinen:
Add validation to those verified personal information fields that were not covered in #322.

Note that `email` field is not touched in this PR as it's obsolete and should be removed. Any way, it's an `EmailField` so there's some validation in place.


profiles/tests/test_migrations.py
```diff
@@ -146,3 +146,29 @@ def verify_migration(apps):
         create_data,
         verify_migration,
     )
+
+
+def test_verified_personal_information_shorter_text_fields_migration(migration_test_db):
```
mikkokeskinen, line 151: This is all well and good, but do we need to verify Django doing migrations correctly? As there seems to be no custom operations in the migration.

Maybe because they are encrypted fields?

profiles/models.py
```diff
@@ -279,6 +280,7 @@ class VerifiedPersonalInformation(ValidateOnSaveModel, NullsToEmptyStringsModel)
         max_length=4,
```
mikkokeskinen, line 282: If it's always exactly 3 digits, could the field have maximum length of 3 too?

mikkokeskinen (approved):
Looks good! Added two observations but nothing that really matters.

profiles/tests/test_migrations.py
akikoskinen: I wrote this test because I thought custom migration code would be needed (to shorten any too long values in the database). But when I run this test, I noticed that it didn't fail and Django actually did the shortening automatically. Which was nice but a surprise to me. So no custom code was needed, but I still left the test be, perhaps to ensure that Django keeps on doing what it did.

I wouldn't say we have to verify Django doing migrations correctly, so in that sense this test is needless. I could also remove this test. Perhaps it has served its purpose, which was to educate me about Django's capabilities :smile:

profiles/models.py
akikoskinen: For consistency that would make sense. I'll fix it.

profiles/tests/test_migrations.py
akikoskinen: Removed that test.

profiles/models.py
akikoskinen: Changed `max_length` to 3.

mikkokeskinen (approved):
👍

PR #328, opened 2021-05-11
Add .python-version file specifying the required Python version

akikoskinen:
The Dockerfile already specifies the Python version this project is built for. This new file is currently mainly meant for Dependabot, so that it can use the correct Python version when it creates dependency update pull requests.

https://github.com/dependabot/dependabot-core/issues/1455


mikkokeskinen (approved):
Pity that it's same as the pyenv version file. Because those kind of deployment-specific files shouldn't be added to the repository. But I guess we have no choice if dependabot doesn't work without it.

akikoskinen:
> Pity that it's same as the pyenv version file.

Yep. I don't know if that's intentional or by accident.

> Because those kind of deployment-specific files shouldn't be added to the repository.

There's also Dockerfile, which defines deployment related things. But you are right, it would be good to not have deployment related things in the same repository.

> But I guess we have no choice if dependabot doesn't work without it.

Another choice is to stop using dependabot in this repository, or prevent it from creating PRs. That's a completely viable option too.

akikoskinen:
> > But I guess we have no choice if dependabot doesn't work without it.
> 
> Another choice is to stop using dependabot in this repository, or prevent it from creating PRs. That's a completely viable option too.

After thinking about this more, should we just tune dependabot not to do those PRs? I suppose we can't teach it to fully make the dependency updates correctly (including using the correct operating system, architecture etc.). We can still have dependabot alerts enabled if those are deemed useful.

mikkokeskinen:
I guess it's up to @tituomin to decide?

tituomin:
> After thinking about this more, should we just tune dependabot not to do those PRs? I suppose we can't teach it to fully make the dependency updates correctly (including using the correct operating system, architecture etc.). We can still have dependabot alerts enabled if those are deemed useful.

dependabot is just an extra, not the primary way we track these dependencies, so if it causes problems, let's tune it thusly.

akikoskinen:
> > After thinking about this more, should we just tune dependabot not to do those PRs? I suppose we can't teach it to fully make the dependency updates correctly (including using the correct operating system, architecture etc.). We can still have dependabot alerts enabled if those are deemed useful.
> 
> dependabot is just an extra, not the primary way we track these dependencies, so if it causes problems, let's tune it thusly.

I vote that the "security updates" (the PRs dependabot makes) should be disabled. [Here's how](https://docs.github.com/en/code-security/supply-chain-security/configuring-dependabot-security-updates#managing-dependabot-security-updates-for-your-repositories) (needs somebody with enough privileges to do that).

The dependabot alerts could still be helpful, but I don't have access to them either, so currently I can't really say are they good or bad.

tituomin:
> > > After thinking about this more, should we just tune dependabot not to do those PRs? I suppose we can't teach it to fully make the dependency updates correctly (including using the correct operating system, architecture etc.). We can still have dependabot alerts enabled if those are deemed useful.
> > 
> > 
> > dependabot is just an extra, not the primary way we track these dependencies, so if it causes problems, let's tune it thusly.
> 
> I vote that the "security updates" (the PRs dependabot makes) should be disabled. [Here's how](https://docs.github.com/en/code-security/supply-chain-security/configuring-dependabot-security-updates#managing-dependabot-security-updates-for-your-repositories) (needs somebody with enough privileges to do that).
> 
> The dependabot alerts could still be helpful, but I don't have access to them either, so currently I can't really say are they good or bad.

Done.

akikoskinen:
Let's not merge this as dependabot would still not be 100% correct with PRs. Dependabot won't be making any PRs any more.

PR #329, opened 2021-05-14
Validate input data for verified personal information addresses early in GraphQL calls

akikoskinen:
This PR changes how input data for verified personal information addresses is validated in the `createOrUpdateProfileWithVerifiedPersonalInformation` mutation. The validation is now done earlier in the mutation handling, even before entering the [`mutate`](https://github.com/City-of-Helsinki/open-city-profile/blob/90ad8203de214a7b30947cf9b4171444ede42643/profiles/schema.py#L909) method.

The implementation uses a [fork of the graphene-validator](https://github.com/City-of-Helsinki/graphene-validator/pull/2) library. ~The needed changes for that library aren't yet merged, so this PR currently references a temporary branch in the fork. That reference needs to be changed before this PR can be merged. **This is the reason why this PR is currently in draft state.**~ **Update:** the needed changes in graphene-validator fork have been merged and this PR is changed to reference the correct branch. This PR is no longer a draft.


profiles/schema.py
```diff
@@ -770,6 +773,11 @@ def mutate_and_get_payload(cls, root, info, **input):
         return CreateProfileMutation(profile=profile)
 
 
+@field_validations(
```
mikkokeskinen, line 776: I was thinking about if the decorator style is a bit too indirect. I mean we have the class right here. Maybe we could write the functions used by the graphene validator by ourselves. Something like:

```python
class VerifiedPersonalInformationAddressInput(graphene.InputObjectType):
    street_address = graphene.String(
        description="Street address with possible house number etc. Max length 1024 characters."
    )
    postal_code = graphene.String(
        description="Postal code. Max length 1024 characters."
    )
    post_office = graphene.String(
        description="Post office. Max length 1024 characters."
    )

    @staticmethod
    def validate_street_address(value):
        return model_field_validation(EncryptedAddress, "street_address", value)

    @staticmethod
    def validate_postal_code(value):
        return model_field_validation(EncryptedAddress, "postal_code", value)

    @staticmethod
    def validate_post_office(value):
        return model_field_validation(EncryptedAddress, "post_office", value)
```

Then the only special thing would be the `model_field_validation` function. Which could be something simple like:

```python
def model_field_validation(model_class, field_name, value):
    field = model_class._meta.get_field(field_name)

    try:
        return field.clean(value, model_class())
    except DjangoValidationError as e:
        raise _multi_validation_error_from_django_validation_error(e)
```

What do you think?

mikkokeskinen (changes requested):
Looking pretty good, but wanted to raise a question.

profiles/schema.py
akikoskinen: That's a possibility too. Having the decorator available doesn't mandate that it needs to be used or that all fields in an input object type should be included in it. It's still ok to write one's own `validate_field_name` function and it can do something special. I wrote the decorator as a helper shorthand for the case that I thought would be quite a common one.

Looking at your alternative code does reveal though, that writing the validator functions by hand isn't that much more code compared to using the decorator so it is a valid question does the decorator really produce value. Perhaps not that much.

I'm starting to lean towards your simpler solution. Let's try it out.

akikoskinen: Updated the last commit, now without the decorator.

utils/decorators.py
```diff
@@ -0,0 +1,50 @@
+import functools
+from typing import Any, Iterable, Mapping
+
+from django.core.exceptions import ValidationError as DjangoValidationError
+from graphene_validator.errors import ValidationError
+
+
+class MultiValidationError(ValidationError):
+    def __init__(self, error_details, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
akikoskinen: Removed those.

utils/decorators.py
```diff
@@ -0,0 +1,50 @@
...
+def _field_validator(model, field, value, info, **input):
+    try:
+        return field.clean(value, model)
+    except DjangoValidationError as e:
+        raise _multi_validation_error_from_django_validation_error(e)
+
+
+def field_validations(**kwargs):
+    model_classes = set(v[0] for v in kwargs.values())
+    models = dict((model_class, model_class()) for model_class in model_classes)
```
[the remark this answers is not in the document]
akikoskinen: This decorator is now removed from the updated version, but if it makes a comeback, I'll keep this in mind.

akikoskinen:
Changed the second to last commit to reference the correct branch from graphene-validator fork and marked this PR as ready for review.

mikkokeskinen (approved):
Good job! 👍

PR #330, opened 2021-05-14
Dependency updates

akikoskinen:
Bundled some dependency updates into the same PR, as the commits themselves are really small. Also removed `coreapi` dependency since I can't find a place in this project where it would be used any more.


akikoskinen:
This should make @tituomin happy :smile:

tituomin (approved):

PR #331, opened 2021-05-19
Validate top level verified personal information early in the GraphQL call

akikoskinen:
This PR does for verified personal information top level data the same thing that #329 did for the VPI addresses: input data is validated even before the `mutate` method is entered. Also fixes documentation.


mikkokeskinen (approved):
LGTM 👍

PR #332, opened 2021-05-26
Allow adding some randomness to audit event output file name

akikoskinen:
Any "X" characters will be replaced by a random character or digit in the audit event output file name.

HP-660

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ functions of Venepaikka service.
 
 Open city profile is implemented using Django and it provides a GraphQL API.
 
+## Configuration
+
+See [docs/config.adoc](docs/config.adoc).
 
 ## Development with [Docker](https://docs.docker.com/)
 
--- a/open_city_profile/settings.py
+++ b/open_city_profile/settings.py
@@ -337,6 +337,17 @@ AUDIT_LOGGING_ENABLED = env.bool("AUDIT_LOGGING_ENABLED")
 AUDIT_LOG_FILENAME = env("AUDIT_LOG_FILENAME")
 
 if AUDIT_LOG_FILENAME:
+    if "X" in AUDIT_LOG_FILENAME:
+        import random
+        import re
+        import string
+
+        system_random = random.SystemRandom()
+        char_pool = string.ascii_lowercase + string.digits
+        AUDIT_LOG_FILENAME = re.sub(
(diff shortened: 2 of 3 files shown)


open_city_profile/settings.py
```diff
@@ -337,6 +337,19 @@ def _list_to_string(value):
...
+        import string
+
+        system_random = random.SystemRandom()
+        char_pool = string.ascii_lowercase + string.digits
+        replaced = ""
+        for char in AUDIT_LOG_FILENAME:
+            if char == "X":
+                char = system_random.choice(char_pool)
+            replaced += char
+        AUDIT_LOG_FILENAME = replaced
```
mikkokeskinen, line 351: I know this is a personal preference, but manual per-character iteration feels a bit off  to me. I would replace this with something like

```python
AUDIT_LOG_FILENAME = re.sub('X', lambda x: system_random.choice(char_pool), AUDIT_LOG_FILENAME)
```

README.md
```diff
@@ -22,6 +22,7 @@ functions of Venepaikka service.
 
 Open city profile is implemented using Django and it provides a GraphQL API.
 
+The service can be configured in many ways. See [separate document](docs/config.adoc).
```
mikkokeskinen, line 25: Suggestion:

```
## Configuration

See [docs/config.adoc](docs/config.adoc).
```

mikkokeskinen (commented):
Good feature. Added two non-essential comments.

open_city_profile/settings.py
akikoskinen: I considered that but discarded it because using regexp for searching a single character feels a bit overkill. But I definitely would like a shorter code for this. I wish there was something like this available:

```python
AUDIT_LOG_FILENAME = replace_all('X', lambda x: system_random.choice(char_pool), AUDIT_LOG_FILENAME)
```

that wouldn't use regexps but just single character comparison. But alas I couldn't find such a thing from the standard library.

But then again, this code is executed in the startup of the program, so there's probably no worrying about using regexp for this either. I'll shorten this code, one way or another.

README.md
akikoskinen: Makes sense.

mikkokeskinen (approved):
👍

PR #334, opened 2021-06-07
Release 1.5.0

akikoskinen:
### New:

- Require service connection while accessing/modifying Profiles in any way. This means that also when a user is requesting their own Profile into some service with the `myProfile` GraphQL query, there needs to be a service connection to that service.
- A service can be configured to behave as if every Profile has a service connection to it. This is called an _implicit connection_.
- Profiles can be searched (with the `profiles` GraphQL query) by verified personal information first and last names and national identification number
- Proper (and stricter) input validation in `createOrUpdateProfileWithVerifiedPersonalInformation` mutation.
- Audit event restructuring and improvements
- Updated dependencies.

### Deprecated/obsoleted:

- All `serviceType` GraphQL arguments are obsolete. They are still in the GraphQL schema, but they have no effect whatsoever.

### Removed:

- Rest of Youth membership code removed


klempine (approved):
LGTM

klempine (approved):
LGTM!

PR #341, opened 2021-06-30
Make verified personal information email obsolete

akikoskinen:
The `email` field is still in the GraphQL schema but it has no effect whatsoever. The column is removed from the database so no email data is stored any more.


tituomin (approved):
It's nice that you have deprecated this by the book, but since this hasn't been published anywhere outside our team yet, it might not be strictly necessary. But I guess / hope it would surface problematic clients of our own...

PR #342, opened 2021-08-02
Remove old authenticator

akikoskinen:
Removed the "old" system that was based on `django-heluser`'s [ApiTokenAuthentication](https://github.com/City-of-Helsinki/django-helusers/blob/master/helusers/_oidc_auth_impl.py#L18), which required the `drf-oidc-auth` library. The `drf-oidc-auth` dependency was removed as well. The first commit contains all this.

After removal of the old authenticator system, the `django-graphql-jwt` library wasn't used for much any more: only a few simple decorators were still used. Since that's not really a good reason to keep a full library as a dependency, implemented those decorators within the project code. The rest of the commits in this PR deal with this change.


open_city_profile/decorators.py
```diff
@@ -56,6 +56,10 @@ def permission_checker(context):
     return permission_checker
 
 
+"""Decorator for checking that the user is logged in"""
```
mikkokeskinen, line 69: It's not possible to add a docstring to a variable. Even if it was, it would have to come on the next line, not previous.

This and the next docstring should therefore be regular comments.

open_city_profile/decorators.py
```diff
@@ -19,23 +19,45 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
-def login_and_service_required(function):
-    """Decorator for checking that the user is logged in and service is known"""
+def _build_decorator(*test_funcs):
+    def decorator(function):
+        @wraps(function)
+        @context_helper(function)
```
mikkokeskinen, line 25: Now that the context_helper is used only here, I would maybe consolidate it and _build_decorator. Could be more easy to follow the flow.

mikkokeskinen (commented):
Good that we can remove the other authenticator!

Added two minor comments.

open_city_profile/decorators.py
akikoskinen: Ah, silly me. Making the string as documentation for the decorator could still be useful. That can be achieved at least like this:

```python
login_required = _build_decorator(_require_authenticated)
login_required.__doc__ = "Decorator for checking that the user is logged in"
```

Does this look ok, or is there a better way?

open_city_profile/decorators.py
akikoskinen: Good point.

open_city_profile/decorators.py
mikkokeskinen: Yeah, that looks ok.

akikoskinen: I did a bigger change for this. I figured that in order for the decorators to be "good citizens", in addition to `__doc__`, they would also need to have correct `__name__` and `__qualname__`. This would have led to:

```python
login_required = _build_decorator(_require_authenticated)
login_required.__name__ = "login_required"
login_required.__qualname__ = "login_required"
login_required.__doc__ = "Decorator for checking that the user is logged in"
```

That's not that simple any more and the function name is repeated six times.

So these new decorators aren't built by assigning to a variable any more, but instead they are real functions introduced with `def`. There is a helper decorator that they use. A docstring within the function works as expected.

mikkokeskinen (approved):
👍

PR #344, opened 2021-08-04
Check sanitizer config in CI

akikoskinen:
As a step in CI builds, check that sanitizer configuration is up-to-date.


mikkokeskinen (approved):
Great addition!

PR #345, opened 2021-08-06
PV-177 turn AddressNode into an entity that can be extended by other services

sam-hosseini:
Hi!

We're building parking-permits and we're using open-city-profile behind a gateway.

I'd like to be able to extend `AddressNode` in our service and add fields that come from KMO service (like getting the coordinates of an address, which zone it is in, etc).

Because at the moment the `AddressNode` does not have a `@key` field, I cannot extend this type in our service.

This PR adds the `@key` field so I can extend it.

### DIFF
--- a/profiles/tests/test_gql_federation.py
+++ b/profiles/tests/test_gql_federation.py
@@ -3,8 +3,8 @@ from graphql_relay import to_global_id
 from guardian.shortcuts import assign_perm
 
 from open_city_profile.tests.asserts import assert_match_error_code
-from profiles.schema import ProfileNode
-from profiles.tests.factories import ProfileFactory
+from profiles.schema import AddressNode, ProfileNode
+from profiles.tests.factories import AddressFactory, ProfileFactory
 from services.tests.factories import ServiceConnectionFactory
 
 GRAPHQL_SDL_QUERY = """
@@ -22,6 +22,19 @@ PROFILE_ENTITY_QUERY = """
             ... on ProfileNode {
                 id
                 firstName
+            }
+        }
+    }
+"""
+
+
+ADDRESS_ENTITY_QUERY = """
+    query ($_representations: [_Any!]!) {
+        _entities(representations: $_representations) {
+            ... on AddressNode {
+                id
+                address
+                postalCode
             }
         }
     }
@@ -67,6 +80,23 @@ def _create_profile_and_variables(with_serviceconnection, service, user=None):
     }
 
     return profile, variables
+
+
+def _create_address_and_variables(with_serviceconnection, service, user=None):
+    profile = ProfileFactory(user=user)
+    address = AddressFactory(profile=profile)
+    if with_serviceconnection:
+        ServiceConnectionFactory(profile=profile, service=service)
+
+    address._global_id = to_global_id(AddressNode._meta.name, address.id)
+
+    variables = {
+        "_representations": [
+            {"id": address._global_id, "__typename": AddressNode._meta.name}
+        ]
+    }
+
+    return address, variables
 
 
 @pytest.mark.parametrize("with_service", (True, False))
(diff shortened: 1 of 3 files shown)


akikoskinen:
I don't see any problem with making the `AddressNode` extendable like this. But just adding the decorator isn't enough. There are some more steps that need to be implemented too.

1. This is a schema change, so schema test needs to be updated (run `pytest --snapshot-update open_city_profile/tests/test_graphql_api_schema.py` to do that)
2. `AddressNode` needs to implement `__resolve_reference`, like [`ProfileNode`](https://github.com/City-of-Helsinki/open-city-profile/blob/413275b584fbeec1e71017067ecc9207808be571/profiles/schema.py#L562-L580) does. This is needed so that the entity is resolved correctly.
3. Access control: the `AddressNode` must be returned as an entity only if the requester has the right to do so. `ProfileNode`'s `__resolve_reference` has access control implemented. I think for the `AddressNode` entity the access control logic should be exactly the same: either the requester is the owner of the entity or the requester has the "can_view_profiles" permission.
4. Tests for the access control. Currently there are [tests for ProfileNode as an entity](https://github.com/City-of-Helsinki/open-city-profile/blob/413275b584fbeec1e71017067ecc9207808be571/profiles/tests/test_gql_federation.py). Something similar is needed for `AddressNode` as entity as well.

akikoskinen:
So far so good. The `__resolve_reference` implementation still needs some tests.

akikoskinen:
Also merge the schema change and schema test change into one commit, so that all tests pass in each commit.

sam-hosseini:
Still need to write the [owner, non-owner, anonymous, staff] access control tests for `AddressNode`. Will try to do that tomorrow.

profiles/tests/test_gql_federation.py
```diff
@@ -4,53 +4,75 @@
...
+                            address
+                            postalCode
+                        }
+                    }
+                }
             }
         }
-    """
+    }
+"""
```
akikoskinen, line 46: Is this how you want to query `AddressNodes` as entities **only**, within a `ProfileNode`? If that's the case, then I think `AddressNode` doesn't have to be an entity at all (doesn't need the `@key` decorator).

If that's not the case, then tests should fetch `AddressNodes` directly as entities. I tried that with the tests and the code seems to deliver exactly that. Here is a patch file [address-node-entity.patch.txt](https://github.com/City-of-Helsinki/open-city-profile/files/6992034/address-node-entity.patch.txt) (the `.txt` suffix is only to persuade Github to accept this file) where I modified the tests to do what I mean.

Let's settle this one question first. Then I have a couple more comments.

sam-hosseini: @akikoskinen That was not the case, I just wasn't so fluent with graphql query syntax and didn't know how to query something on AddressNode directly.

I cannot thank you enough for having written the patch, it made my life so much easier 🙏🏼

I edited the git commit history 👍🏼

profiles/tests/test_gql_federation.py
```diff
@@ -3,54 +3,71 @@
...
+        _entities(representations: $_representations) {
+            ... on AddressNode {
+                id
+                address
+                postalCode
             }
         }
-    """
+    }
+"""
```
akikoskinen, line 41: These can be merged into one query now:

```python
ENTITY_QUERY = """
    query ($_representations: [_Any!]!) {
        _entities(representations: $_representations) {
            ... on ProfileNode {
                id
                firstName
            }
            ... on AddressNode {
                id
                address
                postalCode
            }
        }
    }
"""
```

profiles/tests/test_gql_federation.py
```diff
@@ -3,54 +3,71 @@
...
 
-    executed = anon_user_gql_client.execute(query)
+
+@pytest.mark.parametrize(
+    "schema_type, is_extendable", [("ProfileNode", True), ("AddressNode", True)],
```
akikoskinen, line 45: The `is_extendable` parameter doesn't seem to have any use. Perhaps it can be removed.

akikoskinen (changes requested):
Looking quite nice already. Just a few more comments.

profiles/tests/test_gql_federation.py
akikoskinen: Ok, Github doesn't present this very clearly. The point was, that `PROFILE_ENTITY_QUERY` and `ADDRESS_ENTITY_QUERY` can be replaced with a single `ENTITY_QUERY` that handles both kinds of entities.

sam-hosseini: Makes sense. Done in 53981a6 👍🏼

profiles/tests/test_gql_federation.py
sam-hosseini: Actually `is_extendable` can be used to also test for nodes that are not entities. I've added 2 sample nodes that aren't extendable in 467f9aa to illustrate the usecase. However, I can remove it if you see fit 👍🏼

akikoskinen: I prefer just removing `is_extendable`.

sam-hosseini: Sure. Done in 4683281a67ebe191c7a540b3ba5178d372149de9 👍🏼

akikoskinen: Nice. Could you squash 4683281a67 into 05996262d6 and then I think this would be done.

sam-hosseini: Ok, I have force-pushed all commits from fe68ac7c09485bf29e4e34286cf1edff4660c584 onwards 👍🏼

akikoskinen (approved):

akikoskinen:
Thanks :+1:

sam-hosseini:
Thank you also for the help and all the feedback 🙏🏼

PR #346, opened 2021-08-09
Add first and last name to "create or update profile" GraphQL mutation

akikoskinen:
This changed:

1. `createOrUpdateProfileWithVerifiedPersonalInformation` mutation was duplicated to `createOrUpdateUserProfile`. The mutation with the longer name is deprecated and it doesn't receive any new changes. All development will be made to the mutation with the shorter name.
2. The `verifiedPersonalInformation` input field isn't required in `createOrUpdateUserProfile` any more.
3. The base profile's first and last names can be set with `createOrUpdateUserProfile` mutation.

### DIFF
--- a/open_city_profile/tests/snapshots/snap_test_graphql_api_schema.py
+++ b/open_city_profile/tests/snapshots/snap_test_graphql_api_schema.py
@@ -127,6 +127,16 @@ type CreateOrUpdateProfileWithVerifiedPersonalInformationMutationPayload {
   profile: ProfileWithVerifiedPersonalInformationOutput
 }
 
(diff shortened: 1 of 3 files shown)


akikoskinen:
Access to the new `createOrUpdateUserProfile` mutation is guarded by the same permission as the previous `createOrUpdateProfileWithVerifiedPersonalInformation` was (and still is): "profiles.manage_verified_personal_information". I didn't rename the permission but should it be renamed? If the permission controls access to this mutation _as a whole_, then it's not only about "verified personal information".

mikkokeskinen (approved):
Looks good to me! 👍

PR #347, opened 2021-08-11
Allow setting primary email `verified` flag

akikoskinen:
Adds possibility to control the `verified` flag of the primary email with the `createOrUpdateUserProfile` mutation.

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -1022,8 +1022,9 @@ class CreateOrUpdateUserProfileMutationBase:
 
         profile.emails.exclude(pk=email.pk).filter(primary=True).update(primary=False)
 
-        if not email.primary:
+        if not email.primary or email.verified is not verified:
             email.primary = True
+            email.verified = verified
             email.save()
 
     @staticmethod
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -903,6 +903,9 @@ class VerifiedPersonalInformationInput(graphene.InputObjectType):
 
 class EmailInput(graphene.InputObjectType):
     email = graphene.String(description="The email address.", required=True)
+    verified = graphene.Boolean(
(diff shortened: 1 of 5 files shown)


profiles/schema.py
```diff
@@ -1018,8 +1018,9 @@ def _handle_primary_email(profile, primary_email_input):
 
         profile.emails.exclude(pk=email.pk).filter(primary=True).update(primary=False)
 
-        if not email.primary:
+        if (not email.primary) or (email.verified is not verified):
```
mikkokeskinen, line 1021: I would write this without the parentheses as that's the way it's usually done in python.

```python
if not email.primary or email.verified is not verified:
```

profiles/tests/test_gql_create_or_update_user_profile_mutation.py
```diff
@@ -347,20 +347,25 @@ def test_create_primary_email_for_an_existing_profile(verified, user_gql_client)
 
 
 @pytest.mark.parametrize("old_is_primary", (True, False))
-def test_existing_primary_email_remains_when_trying_to_set_the_same_email_as_a_primary_email(
-    old_is_primary, user_gql_client
+@pytest.mark.parametrize("verified", [None, False, True])
```
mikkokeskinen, line 350: Maybe always use tuples or always use lists in these?

mikkokeskinen (commented):
Good stuff. Added two insignificant code style comments.

profiles/schema.py
akikoskinen: Changed.

profiles/tests/test_gql_create_or_update_user_profile_mutation.py
akikoskinen: Off course. Staying with tuples.

PR #348, opened 2021-08-16
OIDC back channel logout endpoint

akikoskinen:
Update required Django-helusers to version 0.7.0. It provides an [OIDC back channel logout](https://openid.net/specs/openid-connect-backchannel-1_0.html) endpoint implementation. That feature is taken into use with the `HELUSERS_BACK_CHANNEL_LOGOUT_ENABLED` feature flag.


mikkokeskinen (approved):
Looks good to me

PR #349, opened 2021-08-19
Keycloak admin client

akikoskinen:
A simple client that calls Keycloak's admin REST API. It only supports three end points:

- [Fetching data for a single user](https://www.keycloak.org/docs-api/15.0/rest-api/#_getuser)
- [Updating data of a single user](https://www.keycloak.org/docs-api/15.0/rest-api/#_updateuser)
- [Sending a verification email for a user](https://www.keycloak.org/docs-api/15.0/rest-api/#_sendverifyemail)

The authentication uses client credentials grant type, like explained in [this documentation](https://www.keycloak.org/docs/latest/server_development/#authenticate-with-a-service-account).

### DIFF
--- a/utils/keycloak.py
+++ b/utils/keycloak.py
@@ -33,9 +33,9 @@ class KeycloakAdminClient:
                 "client_id": self._client_id,
                 "client_secret": self._client_secret,
             }
-            client_credentials = self._session.post(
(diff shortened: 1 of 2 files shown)


tituomin (approved):
Looks clean!

mikkokeskinen (approved):
👍

PR #351, opened 2021-08-26
Deny getting rid of a Profile's primary email

akikoskinen:
This applies to Profile modifications done via the GraphQL API. If a Profile has a primary email, it can't be modified into a state where it no longer has a primary email.

The beef is in the last commit. All the previous commits are test code refactorings.


tituomin (approved):
Thoroughly tested, indeed!

PR #356, opened 2021-09-16
Validate phone number input in GraphQL

akikoskinen:
Changed `Phone` model so that its `phone` field can't be `null` any more.

Added validation for phone numbers to everywhere in the GraphQL operations where phone numbers can be touched.

### DIFF
--- a/open_city_profile/tests/conftest.py
+++ b/open_city_profile/tests/conftest.py
@@ -154,3 +154,8 @@ def get_unix_timestamp_now():
 @pytest.fixture
 def unix_timestamp_now():
     return get_unix_timestamp_now()
+
+
+@pytest.fixture(params=[None, ""])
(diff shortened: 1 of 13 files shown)


mikkokeskinen (approved):
👍

akikoskinen:
> +1

I realised that this can still be improved and made easier to understand. The second to last commit was changed. It now also documents the new `ExistingProfileInputValidationBase.create_profile` method that sub classes need to implement. Additionally there's now only one version of the `execute_query` method whereas before it had different signatures depending on whether the test class inherited `ProfileInputValidationBase` or `ExistingProfileInputValidationBase` (which received an extra `profile` argument).

mikkokeskinen (approved):
Yeah, even better. 👍

PR #357, opened 2021-09-21
Validate sensitive data in GraphQL mutations

akikoskinen:
There's only the `ssn` field in `SensitiveData` so that's all there is to validate. It is now required that `ssn` is a Finnish personal identification number.


mikkokeskinen (approved):
Looks good.

Though, one might wonder if the sensitive data should remain uneditable using GraphQL.

akikoskinen:
> Though, one might wonder if the sensitive data should remain uneditable using GraphQL.

You mean uneditable _completely_ via GraphQL, or uneditable with `createMyProfile` mutation, which received the editing feature in this PR?

mikkokeskinen:
I don't know. 😊 Just thinking if the field was only meant for staff users or something.

akikoskinen:
> I don't know. Just thinking if the field was only meant for staff users or something.

I don't know either. The ssn is probably given by the profile's owner anyway, so it feels it doesn't make much difference whether that person types it in directly, or some administrator does it.

PR #358, opened 2021-09-21
Fix Mailer message data sanitation configuration

akikoskinen:
The `message_data` database column in tables `mailer_message` and `mailer_messagelog` is NOT NULL. Hence it can't be sanitised to NULL or the data can't be imported. Instead generate a random string of the same length as the original value.


mikkokeskinen (approved):
Good catch. I guess in my tests there were no data on those tables. Should've been more careful.

PR #359, opened 2021-09-21
Verified personal information amr claim check

mikkokeskinen:
This new functionality is used to further secure the access to verified personal information.

"amr" (Authentication Methods Reference) claim is set by the OP to be the name of the authentication backend the user used to authenticate.

If the VERIFIED_PERSONAL_INFORMATION_ACCESS_AMR_LIST-setting is empty, any authentication method grants access.

Refs HP-593

### DIFF
--- a/docs/config.adoc
+++ b/docs/config.adoc
@@ -30,7 +30,7 @@ The following environment variables can be used to affect some standard Django s
 - `EMAIL_URL`: Configures email for Django using https://django-environ.readthedocs.io/en/latest/#email-settings[Django-environ]. Default is "consolemail://".
 - `VAR_ROOT`: Provides a base path for the https://docs.djangoproject.com/en/2.2/ref/settings/#media-root[`MEDIA_ROOT`] and https://docs.djangoproject.com/en/2.2/ref/settings/#static-root[`STATIC_ROOT`] Django settings. `MEDIA_ROOT` will be set to `${VAR_ROOT}/media` and `STATIC_ROOT` to `${VAR_ROOT}/static`. If run with Docker (determined by checking that the project is in an `/app` directory), `VAR_ROOT` defaults to `/var`. Otherwise it defaults to a path called `var` under the project directory.
 
-== Authentication server(s)
+== Authentication and authorization
 
 This project uses https://github.com/City-of-Helsinki/django-helusers[Django-helusers] for doing authentication. The following environment variables correspond directly to Django-helusers settings:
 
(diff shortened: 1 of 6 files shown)


open_city_profile/settings.py
```diff
@@ -68,6 +68,7 @@
     KEYCLOAK_REALM=(str, ""),
     KEYCLOAK_CLIENT_ID=(str, ""),
     KEYCLOAK_CLIENT_SECRET=(str, ""),
+    VERIFIED_PERSONAL_INFORMATION_ACCESS_AMR_LIST=(list, []),
```
akikoskinen, line 71: There is a `docs/config.adoc` file documenting all the settings. This new setting should be added there too.

profiles/tests/test_gql_profiles_query.py
```diff
@@ -194,29 +194,55 @@ def test_staff_user_can_filter_profiles_by_a_field(
...
 
-@pytest.mark.parametrize(
-    "field_name", ["first_name", "last_name", "national_identification_number"]
-)
-def test_staff_user_can_not_filter_profiles_by_verified_personal_information_without_required_permission(
-    field_name, user_gql_client, group, service
+@pytest.mark.parametrize("amr_claim_value", [None, 0, "helsinki_adfs", "foo"])
+@pytest.mark.parametrize("has_needed_permission", [True, False])
+def test_staff_user_filter_profiles_by_verified_personal_information_permissions(
+    has_needed_permission, amr_claim_value, settings, user_gql_client, group, service
```
akikoskinen, line 200: Above this test is a test called `test_staff_user_can_filter_profiles_by_a_field`. Then there was (before this change) a test called `test_staff_user_can_not_filter_profiles_by_verified_personal_information_without_required_permission`. The first one tested cases where filtering succeeds, the latter tested cases where filtering doesn't succeed.

Could that separation of concerns be preserved between these two tests? So the first test would check that filtering works when permissions, settings and amr-claims are all as needed and the second test would check that if any of those is not as needed, then filtering returns no results?

akikoskinen (changes requested):

profiles/tests/test_gql_profiles_query.py
mikkokeskinen: This new function tests all of the cases when searching by the verified personal information. Are you saying we should not parametrize, but separate in two test functions?

akikoskinen: There were two test functions already before. Both of them can be parametrized as needed. I mean that the first test would check all the cases where the filtering returns some profiles (i.e. correct permissions are in place) and the second test would check all the cases where no profiles are returned (i.e. some permission or token claim is missing or incorrect).

akikoskinen: The existing `test_staff_user_can_filter_profiles_by_a_field` also filters by verified personal information, although it doesn't say so in its name.

docs/config.adoc
```diff
@@ -30,18 +30,19 @@ The following environment variables can be used to affect some standard Django s
...
 - `TOKEN_AUTH_ACCEPTED_SCOPE_PREFIX`: Corresponds to `OIDC_API_TOKEN_AUTH. API_SCOPE_PREFIX`. Default is empty.
 - `TOKEN_AUTH_REQUIRE_SCOPE`: Corresponds to `OIDC_API_TOKEN_AUTH. REQUIRE_API_SCOPE_FOR_AUTHENTICATION`. Default is `False`.
 
-The following environment variables configure authentication in other ways:
+The following environment variables configure authentication or authorization in other ways:
 
 - `TOKEN_AUTH_AUTHSERVER_URL`: Sets the "main" authentication server's URL. The URL needs to be exactly what the authentication server reports as its `issuer` value. Default is empty.
 - `ADDITIONAL_AUTHSERVER_URLS`: Sets additional authentication server URLs as a list of strings. JWTs signed by these servers are accepted for authentication. The URLs need to be exactly what the authentication servers report as their `issuer` value. Default is empty list.
+- `VERIFIED_PERSONAL_INFORMATION_ACCESS_AMR_LIST`: Can be used to limit staff users access to verified personal information fields for only those that have authenticated using certain authentication method. If empty, access is not limited. Default is empty list.
```
akikoskinen, line 45: Perhaps this could mention that the authentication method corresponds to the JWT's amr claim.

docs/config.adoc
```diff
@@ -30,18 +30,19 @@ The following environment variables can be used to affect some standard Django s
...
 - `TOKEN_AUTH_ACCEPTED_SCOPE_PREFIX`: Corresponds to `OIDC_API_TOKEN_AUTH. API_SCOPE_PREFIX`. Default is empty.
 - `TOKEN_AUTH_REQUIRE_SCOPE`: Corresponds to `OIDC_API_TOKEN_AUTH. REQUIRE_API_SCOPE_FOR_AUTHENTICATION`. Default is `False`.
 
-The following environment variables configure authentication in other ways:
+The following environment variables configure authentication or authorization in other ways:
 
 - `TOKEN_AUTH_AUTHSERVER_URL`: Sets the "main" authentication server's URL. The URL needs to be exactly what the authentication server reports as its `issuer` value. Default is empty.
 - `ADDITIONAL_AUTHSERVER_URLS`: Sets additional authentication server URLs as a list of strings. JWTs signed by these servers are accepted for authentication. The URLs need to be exactly what the authentication servers report as their `issuer` value. Default is empty list.
+- `VERIFIED_PERSONAL_INFORMATION_ACCESS_AMR_LIST`: Can be used to limit staff users access to verified personal information fields for only those that have authenticated using certain authentication method. (Denoted by the "amr" claim in the authentication token) If empty, access is not limited. Default is empty list.
```
akikoskinen, line 45: Dot should probably be _after_ the parenthesised phrase, not before it?

mikkokeskinen: The most correct could be `...that have authenticated using certain authentication method. (Denoted by the "amr" claim in the authentication token.)` But whatever.

akikoskinen (approved):

PR #360, opened 2021-09-23
Add test and rename variable

akikoskinen:
Just a few small changes that I had done at some point while doing something else.


mikkokeskinen (approved):
👍

PR #361, opened 2021-09-24
Validate profile name fields in GraphQL mutations

akikoskinen:
Includes `Profile` model's `first_name`, `last_name` and `nickname` fields. Maximum length is validated, there's nothing else to check. GraphQL mutations now return a `VALIDATION_ERROR` if a too long value is given.

Also changed how these same fields behave when they are set to `null` (which the GraphQL schema has always allowed). Previously it produced a `GENERAL_ERROR` (coming all the way from the database). Now `null` is turned into an empty string, no error is produced.


profiles/models.py
```diff
@@ -84,9 +84,9 @@ def get_notification_context(self):
 @reversion.register()
 class Profile(UUIDModel, SerializableMixin):
     user = models.OneToOneField(User, on_delete=models.PROTECT, null=True, blank=True)
-    first_name = models.CharField(max_length=255, blank=True, db_index=True)
-    last_name = models.CharField(max_length=255, blank=True, db_index=True)
-    nickname = models.CharField(max_length=32, blank=True, db_index=True)
+    first_name = NullToEmptyCharField(max_length=255, blank=True, db_index=True)
```
mikkokeskinen, line 87: Is there a reason why

```python
first_name = models.CharField(max_length=255, blank=True, db_index=True, default="")
```

is not sufficient?

akikoskinen: The default value is used only if the field isn't set at all. If the field is set to `None`, default is not used and Django tries to store a `null` into the database, which leads to an exception. The `NullToEmptyCharField` turns a value that has been explicitly set to `None`, into an empty string before saving happens.

mikkokeskinen: Ok. But isn't it an error even before the save if the value is `None`?

akikoskinen: I guess that's open to debate.

The reasoning behind this comes from GraphQL's shortcomings. GraphQL uses the same feature for input field optionality and nullness. That's the familiar `!` after a field's type in GraphQL schema. Input field `firstName: String!` is mandatory and it can't be set to `null`. Input field `firstName: String` is optional and if given, `null` is an accepted value (according to the GraphQL schema). There are places in the GraphQL schema where the `firstName` field must be optional, so those use the latter form. As a result, a `null` can be provided as a value to `firstName` via the GraphQL API. Now there are at least a few options how those incoming `null` values should be handled.

1. Treat `null` as an error. Need to document that `null` value is not accepted for `firstName` and if the user still sends one, we deny the request and return an error.
2. Don't treat `null` as an error. Instead take it to mean the same as an empty string and just accept the `null`, silently turning it into an empty string behind the scenes.

I think the latter is friendlier for the end user. No need to document any special case, the GraphQL schema already contains all necessary information. That's why that approach was chosen.

If that's an OK choice, then there are probably also multiple ways to implement this. It could be done close to the GraphQL layer. Or at the model's field level. Or possibly somewhere else. The model field level was chosen, because it's a simple change and affects every input source.

mikkokeskinen: Thanks for the explanation.

Something just feels off when in some cases the validations will handle invalid values and in this case it's the save method correcting errors. But maybe the trade-off here is ok.

akikoskinen: Yes, this is very inconsistent throughout the GraphQL API :disappointed:. I guess in most places giving a `null` input value produces and error. And it's not necessarily even a validation error, but some error coming from the database, which to me indicates a programming error. Those programming errors should be fixed. I would like to go to the direction where `nulls` are accepted as empty input as described earlier.

mikkokeskinen (approved):

tituomin (approved):

PR #362, opened 2021-09-28
Validate email input in GraphQL

akikoskinen:
The email address inputs were already validated before. Improved test coverage to cover all mutations that handle emails. Also refactored the implementation to do validation earlier in the request handling.


mikkokeskinen (approved):
Looks good

PR #365, opened 2021-10-04
Improve profile serialization test coverage

akikoskinen:
The `Profile` serialization output wasn't fully covered by tests. Increased that coverage. Didn't include `subscriptions` as it's an obsolete concept.


mikkokeskinen (approved):
Good idea!

PR #366, opened 2021-10-05
Fix setting service connection creation time

akikoskinen:
The `created_at` field of `ServiceConnection` can't be set at creation time because the field has `auto_now_add` attribute set to `True`. To circumvent that the `created_at` field is now set as a second step after the creation.


mikkokeskinen (approved):
👍

PR #367, opened 2021-10-06
Allow using the user UUID in the Service's GDPR URL

mikkokeskinen:
Additionally allow specifying the exact place where the profile ID or user UUID will be used in the URL by treating the GDPR URL as a template.

The ' and " strings in the help_text are that way because of the black.


services/models.py
```diff
@@ -159,7 +177,7 @@ def delete_gdpr_data(self, api_token: str, dry_run=False):
             data["dry_run"] = True
 
         if self.service.gdpr_url:
-            url = urllib.parse.urljoin(self.service.gdpr_url, str(self.profile.pk))
+            url = self.service.get_gdpr_url_for_profile(self.profile)
```
akikoskinen, line 180: The URL generation could be before the `if` clause. Let the `Service` generate the URL and make decisions only based on the result.

mikkokeskinen: True.

PR #368, opened 2021-10-11
Include verified personal information in downloadMyProfile query

akikoskinen:
Getting VPI requires high enough authentication level, similarly to `myProfile` query.


akikoskinen:
> Is there a compelling reason why the first two commits shouldn't be only one?

No other than producing small commits.

PR #376, opened 2021-11-05
Remove "image" field from Profile 

akikoskinen:
The Profile's `image` field hasn't been working, when used via the GraphQL API. It didn't store any image but just a string. Also according to current plans there will not be an image in the Profile in the first place.

Removed the `image` field from the Profile model. This allowed removing the Pillow dependency as no `ImageFields` are used any more.

In the GraphQL schema all the `image` fields are marked as deprecated and they effectively do nothing.


mikkokeskinen (approved):
Looks good to me.

We should maybe delete the profile_images-directory from the media folder in every environment too.

PR #377, opened 2021-11-09
Check if database contains some obsolete stuff

akikoskinen:
Added a couple of system checks. See commit messages for more details.

### DIFF
--- a/open_city_profile/checks.py
+++ b/open_city_profile/checks.py
@@ -1,0 +1,36 @@
+from django.core.checks import register, Tags, Warning
+from django.db import connections
+
+
+@register(Tags.database)
+def check_obsolete_database_tables(app_configs, **kwargs):
+    KNOWN_UNMANAGED_TABLE_NAMES = {"django_migrations"}
+
+    errors = []
+
+    all_connections = connections.all()
+
+    # Only one database connection is supported
+    if len(all_connections) == 1:
+        conn = all_connections[0]
+
+        existing_table_names = conn.introspection.table_names()
+        expected_table_names = conn.introspection.django_table_names()
+
+        possibly_obsolete_table_names = (
+            set(existing_table_names)
+            - set(expected_table_names)
+            - KNOWN_UNMANAGED_TABLE_NAMES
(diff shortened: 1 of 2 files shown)


open_city_profile/checks.py
```diff
@@ -0,0 +1,59 @@
+from django.contrib.contenttypes.models import ContentType
+from django.core.checks import register, Tags, Warning
+from django.db import connections
+
+
+@register(Tags.database)
+def check_obsolete_database_tables(app_configs, **kwargs):
+    known_unmanaged_table_names = {"django_migrations"}
```
tituomin, line 8: UPPERCASE var name?

akikoskinen: Changed.

mikkokeskinen (approved):
Code looks ok.

Though, in my opinion these should not be checked every time the server starts but with a management command or similar.

akikoskinen:
> in my opinion these should not be checked every time the server starts but with a management command or similar.

I'm curious: why so? Do you mean these shouldn't be using Django's system check framework? If so, why not?

These checks are tagged with the `database` tag, so they get run only with the `migrate` command or if explicitly requested to run.

mikkokeskinen:
Seems I wasn't familiar enough with the Django's system check framework. This makes my point moot:

> For performance reasons, checks are not run as part of the WSGI stack that is used in deployment. If you need to run system checks on your deployment server, trigger them explicitly using check.

akikoskinen:
> Seems I wasn't familiar enough with the Django's system check framework. This makes my point moot:
> 
> > For performance reasons, checks are not run as part of the WSGI stack that is used in deployment. If you need to run system checks on your deployment server, trigger them explicitly using check.

Yep, that's one point. But another is that the `migrate` command **is** run each time the container starts. So these checks will be run each time too a container starts. But the `migrate` command itself probably causes a lot more database traffic any way, just to figure out what migrations it needs to install, if any.

It's also possible to _not_ run the `migrate` command each time the container starts. That's just how things are configured currently.

PR #378, opened 2021-11-12
Remove the legal relationship concept

akikoskinen:
This was an old thing of the past that hasn't been in use. It wasn't even revealed via the GraphQL API in any way.

After these changes `django-ilmoitin` dependency is not needed any more. That will be removed later, once there has been an opportunity to run a "zero migration" for `django-ilmoitin` (and actually `django-mailer` as well, which isn't needed any more either).


mikkokeskinen (approved):
Looks good to me

PR #381, opened 2021-11-29
Don't check for obsolete ContentTypes on empty DB

charn:
Running migrations for an empty DB was failing since the table for `ContentType`s does not exist yet.


akikoskinen (approved):
Looks good. I'm not sure whether the logging is even necessary. If there is no content type table in the database, then there are no obsolete content types either :smile:

charn:
I refactored the extra logging I added into a comment. :+1:

PR #382, opened 2021-12-09
Remove production deploy GitHub workflow (master branch)

akikoskinen:
This change is already present in `develop` branch (e078891b1aa7a0). But we can't advance `master` branch yet to that far, so this one commit could be cherry-picked.


tuovinensanttu (approved):
👌

PR #386, opened 2021-12-23
Remove "concepts of interest" concept

akikoskinen:
This was obsolete and not in use.

This PR enables removing the `django-thesaurus` dependency. That's not included yet in this PR in order to give time to make a zero migration for the thesaurus app in deployments.


mikkokeskinen (approved):
Changing old migrations to point to another model looks dangerous. But it seems to work.

I think we should release version 1.0 when most of the things are "ready" and squash migrations to remove all kinds of migration baggage.

akikoskinen:
> Changing old migrations to point to another model looks dangerous. But it seems to work.

It does look fishy, but I couldn't find any other guidance from the internets either.

> I think we should release version 1.0 when most of the things are "ready" and squash migrations to remove all kinds of migration baggage.

Agree. Although there's already version 1.6 so perhaps the new version would be 2.0 or something like that. But that's details.

PR #387, opened 2022-01-03
Remove "divisions of interest" concept

akikoskinen:
This concept is not really in use. It wasn't even revealed via the GraphQL API.

Removing "divisions of interest" allows getting rid of the Django-munigeo dependency. All code references to Django-munigeo have been removed in this PR. The actual dependency is not yet removed in this PR in order to allow a zero migration to be performed for `munigeo`.


mikkokeskinen (approved):
LGTM

PR #388, opened 2022-01-04
Change usages of ugettext_lazy to gettext_lazy

akikoskinen:
The translation functions starting with 'u' are meant to be used for Python 2 compatibility. This project doesn't support Python 2. Also the 'u' prefixed functions will be removed in Django 4.0.


mikkokeskinen (approved):
👍

PR #390, opened 2022-01-05
Next release

akikoskinen:
Proposing to merge commit a434892a86 to `master` branch.

Draft for release notes:

### New:

* The Services' "implicit connection" feature has been replaced with a "Profile service" feature. The implicit connection property could be added to as many Services as wanted. However it had become apparent that it was only really useful for the Profile Service itself. The Profile Service is special in other ways too, e.g. it doesn't provide a GDPR API. Now a single Service can be marked as the "Profile service".

### Deprecated/obsoleted:

* The subscriptions concept has been deprecated. It's still present in the GraphQL schema but it has been deactivated and it doesn't do anything any more.
* The Profile's image field in GraphQL has been marked as deprecated. It has never actually worked so it has always been useless. There is no "image" in Profile so there's no replacement.


akikoskinen:
Right now this PR merges the tip of `develop` branch to `master`. If we manage to do that, then `master` is ahead of `develop`. That means that `develop` could be fast-forwarded to `master`. Then after that `develop` could move forward again with new commits and `master` could be fast-forwarded to `develop` (or any intermediate commit). This would make `develop` and `master` follow a single line of commits and there wouldn't be need to do merge commits any more. I would really like that.

akikoskinen:
> If the projects using profile are ok for the new release then we are good to go?

As far as I know, yes.

PR #392, opened 2022-01-10
Small updates to README

akikoskinen:
Update build badge. Remove obsolete stuff. Change API documentation link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # Open city profile
 
-[![status](https://travis-ci.com/City-of-Helsinki/open-city-profile.svg)](https://github.com/City-of-Helsinki/open-city-profile)
+[![Continuous integration](https://github.com/City-of-Helsinki/open-city-profile/actions/workflows/ci.yml/badge.svg)](https://github.com/City-of-Helsinki/open-city-profile/actions/workflows/ci.yml)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -181,27 +181,10 @@ The tests require a Postgres database to which to connect to. Here's one way to
 * Run tests within the Django container: `docker-compose exec django pytest`.
 
 
-## Issue tracking
```
mikkokeskinen, line 184: These shouldn't be removed. See [City of Helsinki documentation practises](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/OS/pages/7738589197/Dokumentaatiok+yt+nn+t+documentation+practices).

mikkokeskinen (changes requested):

README.md
akikoskinen: Ah yes, link to issue tracking. But the documentation practices also says that under "Links to public information", so I guess the non-public link can be removed?

mikkokeskinen: Yes, I would assume so.

akikoskinen: According to the guidelines publicly accessible environments should also be listed, so I think I'll resurrect (and correct) those too.

mikkokeskinen (approved):

PR #395, opened 2022-01-11
Add missing migrations for #393

akikoskinen:
#393 changed the default ordering for various models. Changing model default ordering makes Django want to create migrations. These migrations were not included in the original PR, but here they are now. These migrations don't have any effect on the database however.


mikkokeskinen (approved):

akikoskinen:
@mikkokeskinen Sairaan nopee :rocket:

PR #396, opened 2022-01-12
Ability to choose translation language on a field

mikkokeskinen:
More information in the first commit message.

Refs HP-1186

### DIFF
--- a/open_city_profile/graphene.py
+++ b/open_city_profile/graphene.py
@@ -1,9 +1,15 @@
 import uuid
+from collections import OrderedDict
 
 import graphene
+import graphql
+from django.conf import settings
 from django.forms import MultipleChoiceField
+from django.utils import translation
 from django_filters import MultipleChoiceFilter
 from graphene_django.forms.converter import convert_form_field
+from parler.models import TranslatableModel
+from parler.utils.context import switch_language
 
 from profiles.loaders import (
     AddressesByProfileIdLoader,
@@ -70,3 +76,56 @@ class GQLDataLoaders:
             self.cached_loaders = True
 
         return next(root, info, **kwargs)
+
+
+HelTranslationLanguageType = graphql.GraphQLEnumType(
+    name="HelTranslationLanguage",
+    values=OrderedDict(
+        {
+            lang[0].upper(): graphql.GraphQLEnumValue(lang[0], description=lang[1])
+            for lang in settings.LANGUAGES
+        }
+    ),
+)
+
+
+HelTranslationDirective = graphql.GraphQLDirective(
+    name="hel_translation",
+    locations=[graphql.DirectiveLocation.FIELD],
+    args={
+        "in": graphql.GraphQLArgument(
+            graphql.GraphQLNonNull(HelTranslationLanguageType),
+            description="The language",
+        ),
+    },
+    description="Get value translated in language",
+)
+
+
+class HelTranslationMiddleware:
+    def resolve(self, next_middleware, root, info, **kwargs):
+        directives = info.field_asts[0].directives
+
+        hel_translation_directives = [
+            directive
+            for directive in directives
+            if directive.name.value == "hel_translation"
+        ]
+        if not hel_translation_directives:
+            return next_middleware(root, info, **kwargs)
+
+        directive = hel_translation_directives[0]
+        in_argument = next(arg for arg in directive.arguments if arg.name.value == "in")
+        language = graphql.value_from_ast(
+            in_argument.value, HelTranslationLanguageType, info.variable_values
+        )
+
+        if isinstance(root, TranslatableModel):
+            # switch_language will change Django language too
+            with switch_language(root, language):
+                return_value = next_middleware(root, info, **kwargs)
+        else:
+            with translation.override(language):
+                return_value = next_middleware(root, info, **kwargs)
+
+        return return_value
--- a/open_city_profile/tests/test_translation_middleware.py
+++ b/open_city_profile/tests/test_translation_middleware.py
@@ -1,0 +1,50 @@
+import graphene
+import pytest
+from django.conf import settings
+from django.utils import translation
+from django.utils.translation import gettext
+from graphene.test import Client
+from graphene_federation import build_schema
+from graphql import specified_directives
+
+from open_city_profile.graphene import (
+    HelTranslationDirective,
+    HelTranslationLanguageType,
+    HelTranslationMiddleware,
+)
+
+
+@pytest.mark.parametrize("language", [lang[0] for lang in settings.LANGUAGES])
+def test_translation_middleware_changes_language_for_field(language):
+    class Query(graphene.ObjectType):
+        text = graphene.String()
+
+        def resolve_text(self, info):
+            # The string "Text" is translated in Django
+            return gettext("Text")
+
+    schema = build_schema(
+        Query,
+        directives=specified_directives + [HelTranslationDirective],
+        types=[HelTranslationLanguageType],
+    )
+    client = Client(schema, middleware=[HelTranslationMiddleware()])
+
+    query = f"""
+        {{
+            text
+            text_translated: text @hel_translation(in: {language.upper()})
+            text_other_directive: text @skip(if: false)
+        }}
+    """
+
+    expected_text = gettext("Text")
+    with translation.override(language):
+        expected_text_translated = gettext("Text")
+
+    result = client.execute(query)
+
+    assert "errors" not in result
+    assert result["data"]["text"] == expected_text
+    assert result["data"]["text_other_directive"] == expected_text
+    assert result["data"]["text_translated"] == expected_text_translated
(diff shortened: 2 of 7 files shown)


open_city_profile/graphene.py
```diff
@@ -70,3 +77,57 @@ def resolve(self, next, root, info, **kwargs):
...
+    description="Get value translated in language",
+)
+
+
+class TranslationMiddleware:
+    def resolve(self, next, root, info, **kwargs):
+        directives = info.field_asts[0].directives
+
+        if not directives:
+            return next(root, info, **kwargs)
```
akikoskinen, line 110: Seems like `directives` is always an array, even if it's empty. In that case **this** early return wouldn't be strictly needed. The execution could continue to the next step and when no `translation_directives` are found, **then** return early.

open_city_profile/graphene.py
```diff
@@ -70,3 +77,57 @@ def resolve(self, next, root, info, **kwargs):
...
+            directive
+            for directive in directives
+            if directive.name == Name("translation")
+        ]
+        if not translation_directives:
+            return next(root, info, **kwargs)
+
+        directive = translation_directives[0]
+        language = graphql.value_from_ast(
+            directive.arguments[0].value, TranslationLanguageType, info.variable_values
```
akikoskinen, line 122: The more correct, albeit longer, way to get the `in` argument could be for example `next(arg for arg in directive.arguments if arg.name.value == "in")`.

open_city_profile/tests/test_translation_middleware.py
```diff
@@ -0,0 +1,50 @@
...
+
+from open_city_profile.graphene import (
+    TranslationDirective,
+    TranslationLanguageType,
+    TranslationMiddleware,
+)
+
+
+@pytest.mark.parametrize("language", [lang[0] for lang in settings.LANGUAGES])
+def test_translation_middleware_changes_language_for_field(user_gql_client, language):
```
akikoskinen, line 18: This test doesn't use the `user_gql_client`.

services/tests/test_services_graphql_api.py
```diff
@@ -246,3 +250,144 @@ def test_normal_user_can_query_own_services_gdpr_api_scopes(
...
+    service.description = "Service description in finnish"
+    service.save()
+
+    service.set_current_language("en")
+    service.title = "Service title in english"
+    service.description = "Service description in english"
+    service.save()
+
+
+@pytest.mark.parametrize("use_variable", [False, True])
```
akikoskinen, line 267: Not sure what this variable on/off parametrization does in this test. To me it seems that it's testing that one can supply values as variables to a GraphQL query. That doesn't seem too relevant with the subject of this test.  It would simplify this test if this parametrization was removed.

services/tests/test_services_graphql_api.py
```diff
@@ -246,3 +250,144 @@ def test_normal_user_can_query_own_services_gdpr_api_scopes(
...
+@pytest.mark.parametrize("language", ["EN", "FI"])
+def test_service_title_translation(
+    user_gql_client, service_factory, use_variable, language
+):
+    profile = ProfileFactory(user=user_gql_client.user)
+    service = service_factory(service_type=ServiceType.BERTH)
```
akikoskinen, line 273: Service type is deprecated. They shouldn't be used in tests unless the point of the test is to check that they still exist. Here I suppose that's not the case.

Instead of using `service_factory` here, you can just request the `service` fixture.

services/tests/test_services_graphql_api.py
```diff
@@ -246,3 +250,144 @@ def test_normal_user_can_query_own_services_gdpr_api_scopes(
...
+                                "description": expected_description,
+                            },
+                        }
+                    }
+                ]
+            }
+        }
+    }
+
+    assert dict(executed["data"]) == expected_data
```
akikoskinen, line 335: The cast to `dict` is unnecessary.

open_city_profile/graphene.py
```diff
@@ -70,3 +77,57 @@ def resolve(self, next, root, info, **kwargs):
...
+        {
+            lang[0].upper(): graphql.GraphQLEnumValue(lang[0], description=lang[1])
+            for lang in settings.LANGUAGES
+        }
+    ),
+)
+
+
+TranslationDirective = graphql.GraphQLDirective(
+    name="translation",
```
akikoskinen, line 94: The GraphQL spec [recommends prefixing custom directive names](http://spec.graphql.org/October2021/#note-39239) with something. So what would a good prefix be? `hel_`? `helsinki_`?

akikoskinen (changes requested):
Overall very nice :+1: A few tiny nitpicks.

open_city_profile/graphene.py
```diff
@@ -70,3 +77,57 @@ def resolve(self, next, root, info, **kwargs):
...
+    def resolve(self, next, root, info, **kwargs):
+        directives = info.field_asts[0].directives
+
+        if not directives:
+            return next(root, info, **kwargs)
+
+        translation_directives = [
+            directive
+            for directive in directives
+            if directive.name == Name("translation")
```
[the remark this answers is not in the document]
mikkokeskinen: That how I had it initially. But I think it's clearer to use the proper class. But of course can change if you think it's necessary.

open_city_profile/graphene.py
mikkokeskinen: True. I like to return early if nothing to be done. But I can remove it.

open_city_profile/graphene.py
mikkokeskinen: The arguments are validated so I don't think either way is more correct?

services/tests/test_services_graphql_api.py
mikkokeskinen: I had it as separate test, but earlier many test have been frowned upon, so I merged the tests. can of course separate them again.

open_city_profile/tests/test_translation_middleware.py
mikkokeskinen: Nice catch.

open_city_profile/graphene.py
mikkokeskinen: Ah. Didn't catch that. Looked at many custom directives and none used prefix. But yeah we should.

I suppose `hel_` could be ok?

open_city_profile/graphene.py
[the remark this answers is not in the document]
akikoskinen: The `Name` class also contains a `loc` attribute. Although it seems they aren't sure [whether that should affect the equality or not](https://github.com/graphql-python/graphql-core-legacy/blob/master/graphql/language/ast.py#L881-L888)...

open_city_profile/graphene.py
akikoskinen: I mean "the correct logic". The correct logic is something like "the argument called `in` specifies the language". Not "the first argument specifies the language". GraphQL has only named arguments, not positional. Here the difference is of course small, because there is currently only one argument and it's mandatory. But the current implementation might fail if ever there will be another argument for the directive.

services/tests/test_services_graphql_api.py
akikoskinen: I meant that you don't have to test at all whether variables can be supplied for a GraphQL query. We can assume that one can. It's something that Graphene (or graphql-core) should handle.

open_city_profile/graphene.py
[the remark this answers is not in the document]
mikkokeskinen: Yeah, maybe it's not designed to work in these kinds of equality checks. I'll change it.

open_city_profile/graphene.py
akikoskinen: Short and concise :+1:

services/tests/test_services_graphql_api.py
mikkokeskinen: It has to be handled in the middleware (I use the `value_from_ast method`) and I didn't at first and noticed only later when tried to use a variable.

I thought it's an error someone could make again later.

akikoskinen: Ah, I didn't realize that connection. Then it's good.

open_city_profile/graphene.py
mikkokeskinen: Yeah, I guess that's true. I'll change it.

Just noticed that Graphene uses built-in shadowing `next` name as a parameter in their middlewares. I think I'll change that to something else.

services/tests/test_services_graphql_api.py
akikoskinen: There are two translatable fields in Service: title and description. What if one is translated with a variable and the other with a direct input? Would not need that `use_variable` parametrization.

mikkokeskinen: I wanted to leave the other without directive. But I can add an alias as a third.

akikoskinen (commented):
This all looks good now. How should the translatable fields be documented?

akikoskinen (approved):

PR #397, opened 2022-01-12
Update django to 3.2.11

akikoskinen:
Some other dependencies needed updates too so that they support the 3 series of Django. Changed code to be compliant with the new Django and dependencies versions.


mikkokeskinen (approved):
Looks good to me

PR #400, opened 2022-01-28
Another take on field translation

mikkokeskinen:
The directive for language choosing (ticket HP-1186) wasn't good idea because Apollo federation mandates that the same directive must be present in all of the services that are part of the federation.

This PR reverts the directive commits and implements a per-field argument named "language" which can be used to tell which language to use.

Refs HP-1186
Refs HP-1234


akikoskinen:
Didn't go through this very much yet. But I noticed that the `Language` GraphQL enum is now used to indicate the translation language too. The previous usage for it has been something like "the language of the profile (owner?)" or "the customer's desired service language". The set of options for these two purposes aren't necessarily the same. They are _currently_ the same, but it's not guaranteed that it will stay like that. I wonder if it would be safer to create a dedicated enum for this translation purpose, as it was with the `hel_translation` directive solution.

mikkokeskinen:
Yeah, that's a point we discussed earlier. We could make a new enum for the translations, but it would be exact copy of the existing enum with a different name.

Maybe we could just leave it as is and worry about it when we need to diverge the two language sets, if ever?

akikoskinen:
Currently the set of values would be the same in both, yes. If they were ever to diverge, it would need deprecation in the API, supporting both old and new versions for some time, eventually removing the old style. In other words PITA. Having two enums feels like a small price to pay to avoid that.

mikkokeskinen:
I don't think it will ever happen, but I'll change it.

open_city_profile/graphene.py
```diff
@@ -1,16 +1,15 @@
...
 from django.forms import MultipleChoiceField
-from django.utils import translation
 from django_filters import MultipleChoiceFilter
+from graphene_django import DjangoObjectType
 from graphene_django.forms.converter import convert_form_field
+from graphene_django.types import ALL_FIELDS
 from parler.models import TranslatableModel
-from parler.utils.context import switch_language
 
+from open_city_profile import settings
```
akikoskinen, line 12: Why isn't `settings` imported from `django.conf`, as is customary?

open_city_profile/tests/settings.py
```diff
@@ -2,6 +2,8 @@
 
 USE_X_FORWARDED_FOR = True
 
+INSTALLED_APPS += ["open_city_profile.tests.app"]  # noqa
```
akikoskinen, line 5: I think it would be good to document what check the `noqa` comment is concerning. I guess in this case it's F405. So the comment would be `# noqa: F405`.

open_city_profile/tests/test_django_parler_object_type.py
```diff
@@ -0,0 +1,114 @@
...
+    schema = Schema(query=Query)
+
+    query = """
+        query {
+            translatedModel {
+                nonTranslatedField
+                translatedFieldInFinnish: translatedField(language: FI)
+                translatedFieldInEnglish: translatedField(language: EN)
+            }
+        }
```
akikoskinen, line 101: Should there be a test that checks that passing the language via a variable also works?

akikoskinen (changes requested):
Very nice reusable solution :+1: A few minor comments.

open_city_profile/graphene.py
```diff
@@ -78,54 +77,106 @@ def resolve(self, next, root, info, **kwargs):
...
+TranslationLanguage = graphene.Enum(
+    "TranslationLanguage", _get_language_choices_from_parler_settings()
+)
+
 
-        return return_value
+class DjangoParlerObjectType(DjangoObjectType):
+    class Meta:
+        abstract = True
+
```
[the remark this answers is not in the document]
mikkokeskinen: It's a Django feature. Has to be abstract because `DjangoObjectType` has a bit of an [issue](https://github.com/graphql-python/graphene-django/issues/423) when trying to subclass it.

open_city_profile/graphene.py
mikkokeskinen: Seems like autoimport fail. I'll fix it.

open_city_profile/tests/settings.py
mikkokeskinen: That style is used in 2/11 cases in the project. But I can add it here.

open_city_profile/tests/test_django_parler_object_type.py
mikkokeskinen: It's handled by graphene in this case. We don't need to handle it so no need to test it separately.

open_city_profile/graphene.py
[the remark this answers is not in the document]
akikoskinen: Aha. It seems to be a graphene-django feature in this case. But an undocumented one, as are so many other features in that library :facepalm:

open_city_profile/tests/settings.py
akikoskinen: Yep, it's not very common. Burden from the olden times.

akikoskinen (approved):

NikoHelle (approved):
I have no input for this

charn (approved):
LGTM! :+1: :shipit:

PR #403, opened 2022-02-02
Add error code for JWT authentication errors

mikkokeskinen:
The "GENERAL_ERROR" error code returned to the client in case of
an AuthenticationError raised in helusers RequestJWTAuthentication
authenticator was deemed to be too generic.

This commit adds a new error code "JWT_AUTHENTICATION_ERROR" to make
it easier for the client to detect those kinds of errors.

Refs HP-943


akikoskinen (approved):
Looks solid. Just wondering if the error code should not be tied to JWT and be just AUTHENTICATION_ERROR :thinking:

mikkokeskinen:
At this time the error can result only from the RequestJWTAuthentication class.

But yes, It would be worth thinking about if we want to return specific error code or a generic authentication error code. (Especially if we would add a new authentication method(s) in the future)

At least for problem solving purposes it's nice to know see "jwt" in the code in my opinion.

akikoskinen:
Yep. There's also the additional text in GraphQL error that to me feels more like debugging information than something to be shown to the end user or interpreted by any client using the API. That could also give more details when figuring out problem situations...regardless of what granularity the error code gives.

akikoskinen:
But yeah, I already approved so I'm fine with the current solution.

PR #413, opened 2022-06-08
Also audit log contact details

mikkokeskinen:
Refs HP-559


mikkokeskinen:
The 50 microseconds epsilon seems to be too small...

profiles/tests/test_audit_log_to_db.py
```diff
@@ -234,6 +267,78 @@ def test_audit_log_update(live_server, profile_with_related):
...
+                    }
+                }) {
+                    profile {
+                        firstName
+                    }
+                }
+            }
+            """
+        )
+        phone = profile.phones.first()
```
akikoskinen, line 292: This could basically be replaced with

    phone = profile_with_related.related_part

except that there seems to be a bug in constructing the `related_part` member :disappointed:  It's not the related model instance but the related manager. That `related_part` member isn't used elsewhere in the tests so that's probably why this has gone unnoticed. Feel free to fix, if it seems useful. It's also fine as it is currently.

mikkokeskinen: I think it's ok as is to keep it more readable.

akikoskinen (approved):

PR #414, opened 2022-06-10
postgis removed

kanguran:


akikoskinen (commented):
I'm not sure if this needs to be done now. Even though Postgis isn't currently used for anything, it doesn't bring any harm by itself either. Notably, removing these references from the code base doesn't remove the Postgis extension from the _production database_. And that would lead to a discrepancy between development and production environments. I would just leave it be as it is for now.

PR #415, opened 2022-08-18
Update dependencies

mikkokeskinen:
Django had two vulnerabilities, and pycryptodome a couple of minor ones too.

Click in the dev requirements has some too, but it would mean we would need to update Black too because our black version doesn't work with the new Click version.

Updating Black would also mean that ~30 files would get reformatted because of new Black features. Should we update Black or no?

Refs HP-1515


akikoskinen (approved):

akikoskinen:
> Click in the dev requirements has some too, but it would mean we would need to update Black too because our black version doesn't work with the new Click version.
> 
> Updating Black would also mean that ~30 files would get reformatted because of new Black features. Should we update Black or no?

Probably not urgent.

PR #421, opened 2022-09-30
Add new mutation for deleting data from one service

mikkokeskinen:
Three features in the ticket:

- Add ability to dry run profile deletion
- **Ability to delete data from only one service**
- Return which services denied profile deletion and why to the caller

This is the second part.

Refs HP-1525


profiles/schema.py
```diff
@@ -1408,6 +1409,54 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+            raise PermissionDenied(
+                _("You do not have permission to perform this action.")
+            )
+
+        service_connections = profile.effective_service_connections_qs().filter(
+            service__name=input["service_name"]
+        )
+
+        if not service_connections:
+            raise ConnectedServiceDeletionFailedError("Service connection not found")
```
akikoskinen, line 1448: Maybe there should be a dedicated error for this case, where no service connection exists for the named service. Now this same error is used for two different scenarios:

1. Deletion was tried but it failed for some reason (this is the old case)
1. No service connection for the given service name was found (this new case)

My reasoning is that it's possible that these cases need to be distinguished in the UI because the next step the user needs to take is different. In case 1 the user may want to try again or if it keeps failing, contact user support or something like that. In case 2 the user doesn't need to take next steps. The named service doesn't have any data about the user (as far as we know anyway) and that's the end result the user wanted.

profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -569,3 +585,66 @@ def test_api_tokens_missing(user_gql_client, service_1, query_or_delete, mocker)
...
+    user_gql_client.execute(query)
+
+    if dry_run:
+        assert mocked_gdpr_delete.call_count == 1
+        assert [c[2]["dry_run"] for c in mocked_gdpr_delete.mock_calls] == [True]
+        assert ServiceConnection.objects.count() == 2
+    else:
+        assert mocked_gdpr_delete.call_count == 2
+        assert [c[2]["dry_run"] for c in mocked_gdpr_delete.mock_calls] == [True, False]
+        assert ServiceConnection.objects.count() == 1
```
akikoskinen, line 623: This should check that the remaining service connection is for `service_2`.

profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -569,3 +585,66 @@ def test_api_tokens_missing(user_gql_client, service_1, query_or_delete, mocker)
...
 
     assert_match_error_code(executed, MISSING_GDPR_API_TOKEN_ERROR)
+
+
+@pytest.mark.parametrize("dry_run", [True, False])
+def test_user_can_delete_data_from_a_service(
+    user_gql_client, service_1, service_2, mocker, dry_run
+):
+    query = DELETE_SERVICE_DATA_MUTATION_TEMPLATE.substitute(
+        service_name=service_1.name, dry_run=str(dry_run).lower()
```
akikoskinen, line 595: Instead of string template manipulation, it's also possible to just pass input variables to GraphQL client and let it handle them. That also makes it easier to test for example input value defaults: don't give some input value at all and see that the used default is what is expected. See e.g. how `CLAIM_PROFILE_MUTATION` is used in `profiles/tests/test_gql_claim_profile_mutation.py` for an example.

And you can also pass Python booleans as is for boolean input variables :slightly_smiling_face:

profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -569,3 +585,66 @@ def test_api_tokens_missing(user_gql_client, service_1, query_or_delete, mocker)
...
+    assert_match_error_code(executed, CONNECTED_SERVICE_DELETION_FAILED_ERROR)
+
+
+def test_error_when_using_service_delete_with_non_existent_profile(user_gql_client):
+    query = DELETE_SERVICE_DATA_MUTATION_TEMPLATE.substitute(
+        service_name="n/a", dry_run="false"
+    )
+
+    profile = ProfileFactory(user=user_gql_client.user)
+    profile.delete()
```
akikoskinen, line 644: Is the temporary Profile needed for something in this test? I don't get it.

profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -569,3 +585,66 @@ def test_api_tokens_missing(user_gql_client, service_1, query_or_delete, mocker)
...
+
+    executed = user_gql_client.execute(query)
+
+    expected_data = {"deleteServiceData": None}
+    assert dict(executed["data"]) == expected_data
```
akikoskinen, line 649: `executed["data"]` is already a `dict`.

profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -569,3 +585,66 @@ def test_api_tokens_missing(user_gql_client, service_1, query_or_delete, mocker)
...
+def test_error_when_trying_to_delete_data_from_an_unknown_service(user_gql_client):
+    query = DELETE_SERVICE_DATA_MUTATION_TEMPLATE.substitute(
+        service_name="unknown_service", dry_run="false"
+    )
+
+    ProfileFactory(user=user_gql_client.user)
+
+    executed = user_gql_client.execute(query)
+
+    assert_match_error_code(executed, CONNECTED_SERVICE_DELETION_FAILED_ERROR)
```
akikoskinen, line 635: This test could have a Service, with a known name that is used in the query. Just that the Profile doesn't have a connection to that Service (the Profile could even have a connection to _another_ Service for completeness).

akikoskinen (commented):
Quite nice. A few suggestions.

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: Its a copy of existing `test_user_gets_error_when_deleting_non_existent_profile`test. Maybe I'll remove it from both.

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: It's like that in the existing tests. I guess I'll remove it from the other tests too.

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: Isn't that just setting superfluous things up?

open_city_profile/tests/snapshots/snap_test_graphql_api_schema.py
```diff
@@ -232,6 +243,7 @@
   updateMyProfile(input: UpdateMyProfileMutationInput!): UpdateMyProfileMutationPayload
   updateProfile(input: UpdateProfileMutationInput!): UpdateProfileMutationPayload
   deleteMyProfile(input: DeleteMyProfileMutationInput!): DeleteMyProfileMutationPayload
+  deleteServiceData(input: DeleteServiceDataMutationInput!): DeleteServiceDataMutationPayload
```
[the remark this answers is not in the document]
mikkokeskinen: Could be better yes.

profiles/schema.py
mikkokeskinen: Good idea. Added a new exception.

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: Thanks. I didn't remember that style at all. 😄

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: Done.

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: Removed.

profiles/tests/test_profiles_graphql_gdpr_api.py
mikkokeskinen: Done.

profiles/tests/test_profiles_graphql_gdpr_api.py
akikoskinen: The service connection to another Service is probably unnecessary.

In [another comment](https://github.com/City-of-Helsinki/open-city-profile/pull/421#discussion_r984480600) I already suggested a dedicated error for the case where no service connection is found. But there are two reasons why a service connection may not be found:

1. Service with the given name doesn't even exist (this reason is checked in this test)
1. Service with the given name exists, but the Profile has no connection to it

Should these reasons even produce a different error code?

If we think about what the user should do next, I guess both cases are similar (or are they?): the user doesn't need to take any further action. That suggests that one error code for both reasons is enough.

Both reasons could still be tested.

mikkokeskinen: Added a test for the second case.

profiles/schema.py
```diff
@@ -1646,6 +1695,18 @@ class Mutation(graphene.ObjectType):
...
     )
     # TODO: Add the complete list of error codes
+    delete_my_service_data = DeleteMyServiceDataMutation.Field(
+        description="Deletes the data of the profile which is linked to the currently authenticated user from one "
+        "connected service.\n\n"
+        "Requires authentication.\n\nPossible error codes:\n\n"
+        "* `PROFILE_DOES_NOT_EXIST_ERROR`: Returned if there is no profile linked to "
+        "the currently authenticated user.\n"
+        "* `MISSING_GDPR_API_TOKEN_ERROR`: No API token available for accessing the service.\n"
+        "* `SERVICE_CONNECTION_NOT_FOUND`: The user is not connected to the service\n"
```
akikoskinen, line 1705: Small documentation typo: it's SERVICE_CONNECTION_NOT_FOUND_ERROR.

I was also thinking if this should be named in a similar way to the PROFILE_DOES_NOT_EXIST_ERROR, so SERVICE_CONNECTION_DOES_NOT_EXIST_ERROR.

mikkokeskinen: Changed.

akikoskinen (approved):

PR #422, opened 2022-10-06
Service connection GraphQL query

akikoskinen:
Adds a new query that can be used to get a service connection identified by the Profile's User's id (UUID) and any client id of the Service. Requires the "services.view_serviceconnection" permission to execute.


profiles/tests/test_gql_service_connection_with_user_id_query.py
```diff
@@ -0,0 +1,71 @@
...
+                gdprUrl
+                gdprQueryScope
+                gdprDeleteScope
+            }
+        }
+    }
+"""
+
+
+def assert_datetimes_equal(datetime_str, datetime_instance):
```
mikkokeskinen, line 34: Could this be in the open_city_profile.tests.asserts module?

profiles/schema.py
```diff
@@ -1645,10 +1647,13 @@ def resolve_profile_with_access_token(self, info, **kwargs):
...
-        return ServiceConnection.objects.get(
-            profile__user__uuid=kwargs["user_id"],
-            service__client_ids__client_id=kwargs["service_client_id"],
-        )
+        try:
+            return ServiceConnection.objects.get(
+                profile__user__uuid=kwargs["user_id"],
+                service__client_ids__client_id=kwargs["service_client_id"],
+            )
+        except ServiceConnection.DoesNotExist:
```
mikkokeskinen, line 1657: Is there a scenario where we would need to know if the client id exists in the first place? Like can the user somehow end up in Keycloak via a service that is not registered to Profile BE? Or was it already checked in Keycloak?

mikkokeskinen (commented):
Looks good and works nicely. I have to questions though.

profiles/tests/test_gql_service_connection_with_user_id_query.py
akikoskinen: It's a bit special assert because it requires the first argument to be a string and the second to be a datetime. That's why I didn't consider it to be very reusable and didn't include it to the common asserts.

mikkokeskinen (approved):
Looks great!

PR #424, opened 2022-10-14
Add services query

akikoskinen:
A new GraphQL query for receiving general information about services. ~Doesn't require any authentication (should it?).~ Requires `services.view_service` permission.

Also adds a new flag to `ServiceNode`: `requiresServiceConnection`.

### DIFF
--- a/services/tests/test_gql_services_query.py
+++ b/services/tests/test_gql_services_query.py
@@ -1,0 +1,26 @@
+from services.models import Service
+
+QUERY = """
+    query {
+        services {
+            edges {
+                node {
+                    name
+                }
+            }
+        }
+    }
+"""
+
+
+def test_can_query_all_services(service_factory, anon_user_gql_client):
+    service_factory.create_batch(3)
+    # Get services via the model so that they are in the default order
+    services = Service.objects.all()
+
+    expected_service_edges = [{"node": {"name": s.name}} for s in services]
+
+    executed = anon_user_gql_client.execute(QUERY, service=None)
+
+    assert "errors" not in executed
+    assert executed["data"] == {"services": {"edges": expected_service_edges}}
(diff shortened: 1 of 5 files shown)


mikkokeskinen:
I guess we could require that the user has a valid token to get the list.

Another thing I was thinking about: Is it necessary to allow the user to query all the fields in Service. Would it be better to add some kind of PublicServiceNode/SimpleServiceNode/RestrictedServiceNode where the only fields would be name, description and requiresServiceConnection? (If that's all the client needs to know)

akikoskinen:
> I guess we could require that the user has a valid token to get the list.
> 
> Another thing I was thinking about: Is it necessary to allow the user to query all the fields in Service. Would it be better to add some kind of PublicServiceNode/SimpleServiceNode/RestrictedServiceNode where the only fields would be name, description and requiresServiceConnection? (If that's all the client needs to know)

It also needs the `allowedDataFields`. But not the `gdpr*` fields. Instead of introducing a new node type, perhaps just require the `services.view_service` permission?

mikkokeskinen:
services.view_service permission sounds good.

akikoskinen:
Added permission requirement (and rebased).

services/tests/test_gql_services_query.py
```diff
@@ -48,3 +48,33 @@ def test_can_filter_according_to_exact_client_id_of_service(
 
     assert "errors" not in executed
     assert executed["data"] == {"services": {"edges": []}}
+
+
+def test_services_require_service_connections_except_profile_service(
+    profile_service, service, anon_user_gql_client
```
mikkokeskinen: At first I though that service was unused here. Removed it and tested and the test was successful. That is of course correct, but then it doesn't test that the requireServiceConnection is set properly.

Could the testing service be created deliberately in the function body or maybe an assert that there are more than one service? To make it fail if someone is as daft as me.

akikoskinen: Yes, perhaps using ServiceFactory explicitly makes it more clear.

akikoskinen: Changed to use ServiceFactory directly.

Also remembered to add the requirement for elevated privileges to the API documentation.

mikkokeskinen (approved):

PR #426, opened 2022-10-14
Handle AnonymousUser as actor in audit logging

akikoskinen:
**Note**: the target branch is `1.9`, which would branch off from the current `master` branch and `release-1.9.0` tag.

The previous implementation assumed that all users have a `uuid` attribute. But the AnonymousUser doesn't have. Having an AnonymousUser as the actor caused an exception to be raised and that wasn't handled anywhere. As an end result the server responded with a 500 HTTP code.

### DIFF
--- a/profiles/audit_log.py
+++ b/profiles/audit_log.py
@@ -83,10 +83,13 @@ def _get_profile_and_loggables(instance):
 
 
 def _resolve_role(current_user, profile_user_uuid):
-    if profile_user_uuid and current_user and profile_user_uuid == current_user.uuid:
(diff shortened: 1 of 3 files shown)


mikkokeskinen (approved):
Seems to work nicely.

I was wondering if it's a problem that the actor_user_id can now be null. But I guess Logstash and Elasticsearch don't require a value there.

akikoskinen:
> I was wondering if it's a problem that the actor_user_id can now be null. But I guess Logstash and Elasticsearch don't require a value there.

This was tested and a log line went to Elastic without problems. Empty values (that's how Kibana shows them) seem to be OK to all parties involved.

PR #430, opened 2022-11-15
Remove obsolete ClientMutationID from the DeleteMyServiceDataMutation

mikkokeskinen:
The ClientMutationID functionality can be removed from the mutation as it has not yet been in use.

DeleteMyServiceDataMutation tests are removed from the test_profiles_graphql_gdpr_delete_without_results because the mutation can no longer be called without results.


akikoskinen (approved):
Looks good. But better not merge this PR yet, before we are certain that no client is using the `clientMutationId` any more.

mikkokeskinen:
DeleteMyServiceData has not been in production yet, so I don't think anyone could have used it? Except the Profile UI in our test environment and the feature in there is still in progress.

akikoskinen:
Yes, Profile UI is the only client for this I suppose. This is not a major issue, just be careful :smiley:

PR #431, opened 2022-11-16
Exclude service connection to Profile service in GraphQL

akikoskinen:
The profile service doesn't need explicit service connections. Nonetheless, there are some such explicit service connections in the database. Don’t return such service connections from the GraphQL API.

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -561,7 +561,7 @@ class ProfileNode(RestrictedProfileNode):
     )
 
     def resolve_service_connections(self, info, **kwargs):
-        return ServiceConnection.objects.filter(profile=self)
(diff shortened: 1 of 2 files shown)


mikkokeskinen (approved):
Looks good.

Btw. If we don't create service connections for the profile service anymore, would it be nice to mention it in the `effective_service_connections_qs` function. Like "This is just for compatibility with old installations" or something?

akikoskinen:
Currently we are still creating service connections for profile service too. We can consider whether we should stop doing that.

mikkokeskinen:
Ah. Ok. From the commit message I thought there wouldn't always be.

PR #434, opened 2022-11-29
Fix audit log actor role for system users

akikoskinen:
System users' actor role was previously incorrectly resolved as "ADMIN". That's because the system users were indistinguishable from admin users. A `User.is_system_user` field is added in order to be able to identify system users.

**Note:** this PR proposes one way to solve the issue. But I also started thinking if there can be some information in the JWT used for the authentication that would identify system users. I'll investigate that option too.


mikkokeskinen (approved):
Looks good.

Yeah, it could be nice to set the value from the token.

We could maybe also have a system where we just define the allowed sub uuid:s in settings and the authentication system would add the system user automatically on the first use.

mikkokeskinen (changes requested):
Migration should be regenerated as the help_text has changed.

mikkokeskinen:
Also, I seem to have problems migrating down to the previous migration by running `./manage.py migrate users 0002`. Is it just me?

akikoskinen:
I'll investigate. I encountered some problem too, but thought it was because I had renamed the migration file at some point.

akikoskinen:
Meanwhile, here's an example system user's JWT. Doesn't seem to contain anything useful that would directly reveal that this is a system user.

```json
{
    "exp": 1669727155,
    "iat": 1669727095,
    "jti": "c6926753-1d30-4913-8c26-2ddd8bfcc007",
    "iss": "http://keycloak-local:9080/auth/realms/helsinki-tunnistus",
    "aud": "helsinkiprofile-be",
    "sub": "38ec06ae-b9ea-40bd-bafb-fc4dba98409e",
    "typ": "Bearer",
    "azp": "profile-push",
    "session_state": "f698634c-c6a1-4183-9284-80b81b5230b9",
    "scope": "",
    "sid": "f698634c-c6a1-4183-9284-80b81b5230b9"
}
```

akikoskinen:
The problem with django-guardian occurs after a migration, whether it's forward or backwards, if the migration removes any field from the `User` model. After migrations django-guardian tries to ensure that an AnonymousUser instance exists in the database. It [tries to get the AnonymousUser from the database](https://github.com/django-guardian/django-guardian/blob/devel/guardian/management/__init__.py#L32), but for some reason it also tries to fetch the recently removed field (database column). That leads to an exception. The migrations have been done already at this point and they persist in the database, so this doesn't make them fail. It's still an annoying feature.

The AnonymousUser in the database is meant to be used, if one wants to give [some permissions to anonymous users](https://django-guardian.readthedocs.io/en/stable/overview.html#features) as well. I don't think open-city-profile currently needs that feature. The feature can be disabled by [setting the `ANONYMOUS_USER_NAME` setting to `None`](https://django-guardian.readthedocs.io/en/stable/configuration.html#anonymous-user-name). This also [bypasses the problematic functionality after migrations](https://github.com/django-guardian/django-guardian/blob/devel/guardian/management/__init__.py#L40-L44).

It seems however that disabling the feature like this isn't without it's own problems...
https://github.com/django-guardian/django-guardian/issues/714
https://github.com/django-guardian/django-guardian/issues/484

akikoskinen:
Disabled django-guardian's anonymous user support. At least all tests pass, so perhaps we aren't affected by those issues some others have encountered in the same situation.

mikkokeskinen (approved):
Yeah, I guess we don't need to use guardian with anonymous users. Migrations seem to work without emitting errors now.

mikkokeskinen (approved):

PR #435, opened 2022-11-30
HP-1685 Make GDPR delete dry-run request conform to the GDPR API spec

mikkokeskinen:
Moved the dry_run into a query parameter and changed the value to lower case "true". Previously the dry_run parameter was set in the body and the value was "True".

Here is the dry-run part of the current API specification:
```
    - name: dry_run
      in: query
      description: If set to true, the actual removal will not be made...
      required: false
      schema:
        type: boolean
```

Refs HP-1685

### DIFF
--- a/profiles/tests/test_profiles_graphql_gdpr_api.py
+++ b/profiles/tests/test_profiles_graphql_gdpr_api.py
@@ -443,8 +443,8 @@ def test_user_can_dry_run_profile_deletion(
 
     assert service1_mocker.call_count == 1
     assert service2_mocker.call_count == 1
-    assert "dry_run=True" in service1_mocker.request_history[0].text
-    assert "dry_run=True" in service2_mocker.request_history[0].text
+    assert service1_mocker.request_history[0].qs["dry_run"] == ["true"]
+    assert service2_mocker.request_history[0].qs["dry_run"] == ["true"]
     assert Profile.objects.filter(pk=profile.pk).exists()
     assert ServiceConnection.objects.count() == 2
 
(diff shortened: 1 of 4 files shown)


profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -443,8 +443,8 @@ def get_response(request, context):
 
     assert service1_mocker.call_count == 1
     assert service2_mocker.call_count == 1
-    assert "dry_run=True" in service1_mocker.request_history[0].text
-    assert "dry_run=True" in service2_mocker.request_history[0].text
+    assert "dry_run=true" in service1_mocker.request_history[0].url
+    assert "dry_run=true" in service2_mocker.request_history[0].url
```
akikoskinen, line 447: The request_history objects have both `qs` and `query` attributes which could be more appropriate for these asserts. Applies also to a few more similar places in this PR.

profiles/tests/test_profiles_graphql_gdpr_api.py
```diff
@@ -500,7 +500,7 @@ def test_user_tries_deleting_his_profile_but_it_fails_partially(
     ServiceConnectionFactory(profile=profile, service=service_2)
 
     def get_response(request, context):
-        if not request.body or "dry_run" not in request.body:
+        if "dry_run=true" not in request.url:
```
akikoskinen, line 503: The `request` object has both `qs` and `query` attributes which could be more appropriate for this check. Applies also to a few more similar places in this PR.

akikoskinen (changes requested):
A few suggestions for stricter tests.

akikoskinen (approved):

PR #438, opened 2022-12-19
downloadMyProfile query error reporting

akikoskinen:
Previously when it wasn't possible to retrieve data from connected services (because of missing configuration or HTTP errors), the errors were silently ignored and skipped. Now the error situations return an appropriate error in the GraphQL response.


mikkokeskinen (approved):
Looks really good! 👍

PR #439, opened 2023-01-09
OpenAPI specification for GDPR API implementations

akikoskinen:
Brought the specification from internal documentation system and updated it.

Now also includes a live documentation page. Start the open-city-profile server and load `/docs/gdpr-api/` path.

### DIFF
--- a/open_city_profile/static/open-city-profile/gdpr-api-openapi.yaml
+++ b/open_city_profile/static/open-city-profile/gdpr-api-openapi.yaml
@@ -7,7 +7,7 @@ paths:
   /profiles/{id}:
     get:
       summary: Retrieve all of the data related to the given id.
-      description: Retrieve all of the data related to the given id, represented in a tree structure. Depending on the `displayable` parameter, the response may contain additional metadata which can be used when formatting the data for a user.
+      description: Retrieve all of the data related to the given id, represented in a tree structure.
       tags:
       - profiles
       operationId: getProfile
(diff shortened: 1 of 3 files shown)


open_city_profile/static/open-city-profile/gdpr-api-openapi.yaml
```diff
@@ -0,0 +1,293 @@
+openapi: 3.0.1
+info:
```
kanguran, line 2: I check which versions are possible to set here and found this link https://en.wikipedia.org/wiki/OpenAPI_Specification
3.0.1 version is 5 years old.
I tried to set version to 3.1.0 and run it locally but I was shown Error that version must be 3.0.X
So, I think OpenAPI version has to be set to 3.0.3 for now.

akikoskinen: Age of a specification doesn't really matter. It can be decades old, if it's still valid and usable.

https://github.com/OAI/OpenAPI-Specification/releases

OpenAPI version 3.1 has some breaking changes compared to 3.0 so this document might need some changes in order to be compliant.

All the 3.0.x versions are identical when it comes to what the specification specifies, the differences are in wording etc. of the specification. In that sense it doesn't make a difference what patch version of 3.0.x series we have here.

I can still change the OpenAPI version to 3.0.3 though, because why not :smiley:

akikoskinen: Changed.

mikkokeskinen (commented):
I think this is excellent and the technical documentation fits naturally in the swagger docs.

Is there still something to do if the last commit message starts with "WIP"?

akikoskinen:
> I think this is excellent and the technical documentation fits naturally in the swagger docs.
> 
> Is there still something to do if the last commit message starts with "WIP"?

Perhaps just missing a proper commit message and also consideration whether it's good to have all that in a single commit or should it be split somehow.

akikoskinen:
> Perhaps just missing a proper commit message and also consideration whether it's good to have all that in a single commit or should it be split somehow.

I wrote a proper commit message for the last commit, but kept it all in a single commit still.

mikkokeskinen (approved):

PR #440, opened 2023-01-11
Return dedicated error code via GraphQL when Keycloak reports a conflict

akikoskinen:
When updating user data in Keycloak, it could return a 409 HTTP code to inform about conflicting data. Such a conflict may be caused for example when trying to change a user's email to a value that is already used by another user. A new GraphQL error code, `DATA_CONFLICT_ERROR`, is now returned in those cases.

Refs HP-1766


profiles/keycloak_integration.py
```diff
@@ -91,7 +94,12 @@ def send_profile_changes_to_keycloak(instance):
     if email_changed:
         updated_data["emailVerified"] = False
 
-    _keycloak_admin_client.update_user(user_id, updated_data)
+    try:
+        _keycloak_admin_client.update_user(user_id, updated_data)
+    except requests.HTTPError as err:
```
mikkokeskinen, line 99: I'm not a big fan of the fact that the "API" for the Keycloak admin client is just requests exceptions.

Could it handle the requests exceptions internally and maybe raise the DataConflictError itself? What do you think?

akikoskinen: I have been thinking the same thing. The client could definitely communicate errors in some clearer way. Let's do it.

Should I make such change within this PR or create another one for that purpose? I can also change the client first in another PR, and then refactor this PR on top of that.

mikkokeskinen: Making it in this PR works for me.

mikkokeskinen (approved):
Excellent!

PR #442, opened 2023-03-22
When profile data sending to Keycloak fails, keep own database untouched

akikoskinen:
Previously the profile updating logic was like this:

- Modify own database (including closing the transaction)
- Send changed profile data to Keycloak
- If previous step fails, return an error to caller (but own database was already changed and it's left like that)

The logic after this PR is:

- Modify profile models in memory
- Send changed profile data to Keycloak
- If previous step fails, return an error to caller
- If we get here, persist the models to own database

The profile data updating logic needed quite a big change. This is built gradually in the several commits. There are also some other improvements, like better handling of invalid Relay global ids. New tests were also added for cases that were discovered to be untested before.

### DIFF
--- a/open_city_profile/tests/asserts.py
+++ b/open_city_profile/tests/asserts.py
@@ -32,5 +32,5 @@ def assert_match_error_code(response, error_code):
     error = response["errors"][0]
     assert "extensions" in error
     extensions = error["extensions"]
-    assert "code" in extensions
-    assert extensions["code"] == error_code
+    assert "code" in extensions, error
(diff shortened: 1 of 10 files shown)


voneiden:
Closing this. We're approaching this issue primarily from a different angle. bef14e5 can be merged in a separate PR. Branch [HP-1781-continued](https://github.com/City-of-Helsinki/open-city-profile/tree/HP-1781-continued) has the commits not included in #445 rebased waiting for further review.

PR #443, opened 2023-03-30
Application logging settings. Some logging to gdpr query.

akikoskinen:
Almost the same as in the [gdpr-query-logging](https://github.com/City-of-Helsinki/open-city-profile/tree/gdpr-query-logging) branch, but now in proper commits. Differences to the WIP branch are:

- Configure logging for all apps in this project
- Use full logger `name` instead of `module` (which would be the part after the last dot of the `name`) in the log message
- A setting for overwriting the log level


mikkokeskinen (approved):
Looks really good to me!

PR #444, opened 2023-03-31
Add logging to GDPR delete functionality

akikoskinen:
First moved the service data deletion code to `connected_services` module, where the GDPR query code already was and the deletion code also partially. Tests required some adjustment to enable that move.


mikkokeskinen (approved):
👍

PR #446, opened 2023-04-11
Update some dependencies, mostly to fix vulnerabilities

akikoskinen:
- Includes updating Black, which caused a lot of formatting changes in the code.
- Got completely rid of some dependencies, like `future`, instead of updating them to fix vulnerabilities.

Some packages with known vulnerabilities were not updated:

- sentry-sdk: Helsinki doesn't use sentry-sdk in reality (it's not configured in any environment), even though it's included in the project. It has changed quite significantly since the last time it has been updated and looks like it would need a more thorough check than just updating the version number. If Sentry is not used in reality, perhaps it should be removed completely.
- IPython: it's a dev dependency and vulnerability only occurs on Windows.


akikoskinen:
> rosetta error: futex(FUTEX_LOCK_PI_PRIVATE) failure: 35

Perhaps something ARM related? I can build images successfully on x86.

mikkokeskinen (approved):

PR #454, opened 2023-05-25
Allowed data fields configurator

akikoskinen:
There was already before a simple `generate_data_fields` function that inserted some hard coded data for the `AllowedDataFields` model into the database. This PR enhances that function to handle all kinds of changes in the `AllowedDataFields` objects. It also modifies the allowed data fields of a `Service`, if a field for example gets renamed (or actually replaced). The data is not hard coded any more in this repository.

There's also a simple management command (`set_allowed_data_fields`) that can be used to call that configurator function. The management command reads the specification for the fields from stdin.


akikoskinen:
Changed to draft PR as there's currently some problems with this that need more attention.

akikoskinen:
Now this is ready for review. The first two commits were added, changing the nature how the `seed_data` (now called `seed_development_data`) management command works and how it's used.

services/tests/test_utils_generate_data_fields.py
```diff
@@ -0,0 +1,161 @@
...
+):
+    allowed_fields_spec = _get_initial_spec()
+    generate_data_fields(allowed_fields_spec)
+
+    services = ServiceFactory.create_batch(2)
+    for service in services:
+        service.allowed_data_fields.set(AllowedDataField.objects.all())
+
+    for field_index in range(num_fields_to_replace):
+        replacable_field = allowed_fields_spec[field_index]
```
mikkokeskinen, line 125: I think it should be "replaceable"

docker-entrypoint.sh
```diff
@@ -20,10 +20,7 @@ fi
 
 if [[ "$APPLY_MIGRATIONS" = "1" ]] && [[ "$SEED_DEVELOPMENT_DATA" = "1" ]]; then
     echo "Seeding initial development data..."
-    ./manage.py seed_data --development
-elif [[ "$APPLY_MIGRATIONS" = "1" ]]; then
```
mikkokeskinen, line 24: Could some installation somewhere depend on the fact that APPLY_MIGRATIONS also adds the seed data?

mikkokeskinen (commented):
Works nicely. Only spotted a typo.

docker-entrypoint.sh
akikoskinen: I guess that's possible. Perhaps this should be marked as a breaking change in the next release?

services/tests/test_utils_generate_data_fields.py
akikoskinen: Fixed

mikkokeskinen (approved):
👍

PR #460, opened 2023-11-15
Hp 2082 update vulnerable dependencies

AntiAki-M:
Updated dependencies with vulnerabilities or security issues

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -167,7 +167,7 @@ typing-extensions==4.5.0
     #   -c requirements.txt
     #   black
     #   platformdirs
-urllib3==1.26.6
+urllib3==1.26.18
     # via
     #   -c requirements.txt
     #   requests
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ database-sanitizer==1.1.0
     # via django-sanitized-dump
 deprecation==2.1.0
     # via django-helusers
-django==3.2.21
+django==3.2.23
     # via
     #   -r requirements.in
     #   django-admin-sortable
--- a/requirements.txt
+++ b/requirements.txt
@@ -133,7 +133,7 @@ sqlparse==0.4.4
     # via django
 typing-extensions==4.5.0
     # via asgiref
-urllib3==1.26.6
+urllib3==1.26.18
     # via
     #   requests
     #   sentry-sdk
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -114,7 +114,7 @@ requests-oauthlib==1.3.0
     # via -r requirements.in
 rx==1.6.1
     # via graphql-core
-sentry-sdk==0.10.2
+sentry-sdk==0.14.0
```
akikoskinen, line 117: Sentry-sdk **0**.14.0 doesn't fix [the vulnerability](https://www.cve.org/CVERecord?id=CVE-2023-28117) mentioned in the commit message. It's fixed in version **1**.14.0.

But the situation with sentry-sdk is a bit sad. The version in use is very old and getting to a modern version would need some more changes in the code base than just raising the version number. It has been left like this because Sentry isn't configured and is actually not used in our installations.

I think it's OK to leave sentry-sdk to the old version. It must be updated if and when somebody decides that we really should be using Sentry.

requirements-dev.txt
```diff
@@ -167,7 +167,7 @@ typing-extensions==4.5.0
     #   -c requirements.txt
     #   black
     #   platformdirs
-urllib3==1.26.6
+urllib3==1.26.18
```
akikoskinen, line 170: This change (commit 0db6b4db) should be squashed into the same commit that changes the version in `requirements.txt` (b7253d8f2).

requirements.txt
```diff
@@ -133,7 +133,7 @@ sqlparse==0.4.4
     # via django
 typing-extensions==4.5.0
     # via asgiref
-urllib3==1.26.6
+urllib3==1.26.18
```
akikoskinen, line 136: The second link in the commit message points to a wrong CVE number. Should be CVE-2023-4**3**804.

akikoskinen (changes requested):
Nice. Few small suggestions.

requirements.txt
AntiAki-M: Oh, didn't even notice it required a major update, I'll revert the change.

requirements-dev.txt
AntiAki-M: Will squash

requirements.txt
AntiAki-M: Will fix

akikoskinen (approved):

PR #462, opened 2023-12-13
Explain Keycloak style scopes in GDPR API documentation

akikoskinen:
Tunnistamo's and Keycloak's scope styles in the API access tokens differ significantly. The previous explanation covered only Tunnistamo style. Now Keycloak style explanation is there too and it's emphasized that these are two different things.


AntiAki-M (approved):
I guess it's accurate

PR #463, opened 2023-12-27
feat: support new ssn intermediary characters

voneiden:
Ref HP-2087

### DIFF
--- a/profiles/tests/test_validators.py
+++ b/profiles/tests/test_validators.py
@@ -49,6 +49,34 @@ class TestValidateFinnishNationalIdentificationNumber:
     def test_accept_valid_string(self, number):
         assert validate_finnish_national_identification_number(number) is None
 
+    @pytest.mark.parametrize(
+        "intermediary",
+        [
+            "+",
+            "-",
+            "A",
+            "B",
+            "C",
+            "D",
+            "E",
+            "F",
+            "U",
+            "V",
+            "W",
+            "X",
+            "X",
+        ],
+    )
+    def test_accept_valid_intermediary_letters(self, intermediary):
+        assert (
+            validate_finnish_national_identification_number(f"010101{intermediary}1231")
+            is None
+        )
+
+    @pytest.mark.parametrize("intermediary", ["_", "G", "1"])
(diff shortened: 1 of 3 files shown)


profiles/tests/test_models.py
```diff
@@ -489,6 +489,10 @@ def test_ssn(self):
 
         self.fails_validation(sensitive_data)
 
+        for intermediary in "+-ABCDEFUVWXX":
+            sensitive_data.ssn = f"150977{intermediary}5554"
+            self.passes_validation(sensitive_data)
```
akikoskinen, line 494: The validator is most thoroughly tested in `profiles/tests/test_validators.py`. Those tests could be enhanced to include these new accepted characters. These other tests just ensure that the validator is used in the correct places, but they don't need to cover more than one case.

akikoskinen (changes requested):
The commit message could elaborate more what the change is about. Perhaps also a link to the population registry's news.

profiles/tests/test_validators.py
```diff
@@ -49,6 +49,34 @@ def execute_denied_number_test(number):
...
+    )
+    def test_accept_valid_intermediary_letters(self, intermediary):
+        assert (
+            validate_finnish_national_identification_number(f"010101{intermediary}1231")
+            is None
+        )
+
+    @pytest.mark.parametrize("intermediary", ["_", "G", "1"])
+    def test_deny_invalid_intermediary_letters(self, intermediary):
+        self.execute_denied_number_test(f"010101{intermediary}1231")
```
akikoskinen, line 78: This test seems to be the same as the already existing `test_deny_invalid_century_character` test. They could be merged into one.

voneiden: Right-o, very much duplicated. Removed the other one.

nicobav (approved):
Codewise LOKTM!

Building review failed though 🤔

akikoskinen (approved):

voneiden:
Yeah  @nicobav the kolga review env is broken, it can be removed, there's a PR in progress on the matter @ #465

PR #465, opened 2023-12-27
Disable review environment

voneiden:
The review environment has been broken for quite a while.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -30,6 +30,7 @@ jobs:
         uses: andersinno/kolga-build-action@v2
 
   review:
+    if: false
     runs-on: ubuntu-latest
     needs: build
     name: Review


akikoskinen (commented):
How about just removing the whole workflow?

voneiden:
Mm not yet. I don't want to hide any evidence 2 days before the full maintenance responsibility shifts to ratkis. There's also the staging environment and gitlab references that need to be checked by Joku :tm: . If they're on longer in use and have been replaced by platta equivalents, then maybe there's further cleaning up to do in the Kubernetes. 

I suppose you've not used these review and stage environments that kolga spins up at any point, or?

akikoskinen:
> Mm not yet. I don't want to hide any evidence 2 days before the full maintenance responsibility shifts to ratkis. There's also the staging environment and gitlab references that need to be checked by Joku ™️ . If they're on longer in use and have been replaced by platta equivalents, then maybe there's further cleaning up to do in the Kubernetes.
> 
> I suppose you've not used these review and stage environments that kolga spins up at any point, or?

These haven't been used for years, AFAIK. These (try to) deploy to some KuVa infrastructure somewhere. I'm surprised that any of them (the staging workflow at least) have succeeded during the last couple of years.

If [this](https://gitlab.com/City-of-Helsinki/KuVa/github-mirrors/open-city-profile/-/pipelines) is the GitLab pipeline, it seems that it's not used either anymore.

There are no direct replacements. There hasn't been a need for review environments. `develop` branch does get automatically deployed to Platta DEV environment, but that's taken care of in Azure, not in this repository.

charn (approved):
LGTM! :+1: :shipit:

PR #468, opened 2024-02-08
HP-2191 | fix: settings primary email shouldn't fail on duplicate emails

charn:
When a user has updated their email e.g. using `updateMyProfile` mutation, there's no check for duplicate emails. Primary email update coming e.g. keycloak shouldn't fail if there's duplicate emails. Instead, using the first matching email should be enough.


voneiden (approved):
Excellent

PR #469, opened 2024-02-13
HP-2196 | Hide national identification in Django admin

charn:
Instead of showing the actual value a boolean check-mark is displayed.

**Screenshot**
Screenshot from 2024-02-13 17-12-47


voneiden (approved):
:heavy_check_mark:

PR #470, opened 2024-02-14
HP-1787 | fix: do keycloak changes inside a transaction

charn:
In case Keycloak returns an error we need to be able to revert.


danipran (approved):

voneiden (approved):
Seems like an improvement to me.  :+1:

PR #494, opened 2024-04-29
HP-2319 | Check allowed data fields when querying profiles

nicobav:
Adds middleware for checking allowed_data_fields for service when querying profiles
Adds feature flag for the feature.

refs HP-2319


terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

danipran (commented):
The code itself seems fine, just a bit confused about the documentation.

open_city_profile/graphene.py
```diff
@@ -178,3 +180,26 @@ def __init_subclass_with_meta__(
             _meta=_meta,
             **options,
         )
+
+
+class AllowedDataFieldsMiddleware:
+    def resolve(self, next, root, info, **args):
+        try:
+            is_my_profile = (
+                info.operation.selection_set.selections[0].name.value == "myProfile"
```
charn, line 189: I wonder if it's sufficient to only filter myProfile queries. A service (e.g. an admin UI of a service) might be using a different query (e.g. `profiles` and `profile`).

profiles/tests/test_gql_my_profile_query.py
```diff
@@ -419,5 +466,52 @@ def test_normal_user_can_query_their_own_profiles_sensitivedata(user_gql_client)
...
     assert dict(executed["data"]) == expected_data
+
+
+def test_my_profile_results_error_if_querying_fields_not_allowed(
+    user_gql_client, service
+):
+    profile = ProfileFactory(user=user_gql_client.user)
+    ServiceConnectionFactory(profile=profile, service=service)
+
+    query = """
```
charn, line 479: If you change this test e.g. to the following, you are able to the field permission check and get first and last name:

```
    query = """
        {
            _service {
                sdl
            }
            myProfile {
                firstName
                lastName
            }
        }
    """
```

`myProfile` is not at the 0-index but in `info.operation.selection_set.selections[1].name.value`.

profiles/schema.py
```diff
@@ -370,6 +371,46 @@ def filter_by_nin_exact(self, queryset, name, value):
...
+    e.g:
+    allowed_data_fields_map = {
+        "name": ("first_name", "last_name", "nickname"),
+        "personalidentitycode": ("national_identification_number",),
+        "address": ("address", "postal_code", "city", "country_code")
+    }
+
+    `always_allow_fields`: Since connections are not defined in `allowed_data_fields.json` they should be
+    defined here. If the field is connection and the node does not inherit this mixin the data will be available
+    to all services.
```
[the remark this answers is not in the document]
nicobav: > So... what's allowed_data_fields.json? 😅 No mention of it in README. It's a bit confusing for me at least.

Oh actually that's a good question in this github context. Our documentation **_elsewhere_** guides about using json file named like that to import the allowed data fields using command `set_allowed_data_fields`. Which then creates the `AllowedDataField` model instances mapped in the json file.

> Checks for allowed data fields, which means...? 

We need to check from the `AllowedDataField` instances linked to `Service` to which fields the service is allowed to access.

profiles/tests/test_gql_my_profile_query.py
nicobav: very good spot indeed. Changed the logic so it goes through all the selections.

open_city_profile/graphene.py
nicobav: That's true, I added the those `profile` and `profiles` to be checked as well.

nicobav:
I made some changes based on the comments and added the feature to be inside a feature flag as discussed irl.

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

open_city_profile/graphene.py
```diff
@@ -178,3 +183,55 @@ def __init_subclass_with_meta__(
...
+            node = None
+
+        if is_profile_query and getattr(node, "check_allowed_data_fields", False):
+            field_name = to_snake_case(getattr(info, "field_name", ""))
+
+            if not getattr(info.context, "service", False):
+                if settings.ENABLE_ALLOWED_DATA_FIELDS_RESTRICTION:
+                    raise ServiceNotIdentifiedError("Service not identified")
+
+                send_allowed_data_field_sentry_warning(
```
charn, line 221: Maybe a standard logger.warning or logger.error would be enough here.

open_city_profile/graphene.py
```diff
@@ -178,3 +183,55 @@ def __init_subclass_with_meta__(
...
+                    "Allowed data field exception would occur: Service not identified",
+                    None,
+                    field_name,
+                )
+
+            if not node.is_field_allowed_for_service(field_name, info.context.service):
+                if settings.ENABLE_ALLOWED_DATA_FIELDS_RESTRICTION:
+                    raise FieldNotAllowedError("Field is not allowed for service.")
+
+                send_allowed_data_field_sentry_warning(
```
charn, line 231: Maybe a standard logger.warning or logger.error would be enough here.

open_city_profile/graphene.py
```diff
@@ -178,3 +183,55 @@ def __init_subclass_with_meta__(
...
+                    "Allowed data field exception would occur. Field is not allowed for service.",
+                    info.context.service,
+                    field_name,
+                )
+
+        return next(root, info, **args)
```
charn, line 237: Just thinking here. What if `AllowedDataFieldsMixin` was used on the models instead and not on the graphene node classes (so `Profile` and `VerifiedPersonalInformation`).

You could then do something like:

```suggestion
    def resolve(self, next, root, info, **kwargs):
        if getattr(root, "check_allowed_data_fields", False):
            field_name = to_snake_case(info.field_name)

            if not getattr(info.context, "service", False):
                if settings.ENABLE_ALLOWED_DATA_FIELDS_RESTRICTION:
                    raise ServiceNotIdentifiedError("Service not identified")
                
                logger.warning(f"Allowed data field exception would occur: Service not identified")

            if not root.is_field_allowed_for_service(field_name, info.context.service):
                if settings.ENABLE_ALLOWED_DATA_FIELDS_RESTRICTION:
                    raise FieldNotAllowedError("Field is not allowed for service.")

                logger.warning(f"Allowed data field exception would occur. Field '{field_name}' is not allowed for service {info.context.service}.")

        return next(root, info, **kwargs)
```

open_city_profile/graphene.py
```diff
@@ -178,3 +183,55 @@ def __init_subclass_with_meta__(
...
+
+class AllowedDataFieldsMiddleware:
+
+    def _is_profile_query(self, info):
+        return any(
+            selection.name.value in ("myProfile", "profile", "profiles")
+            for selection in info.operation.selection_set.selections
+        )
+
+    def resolve(self, next, root, info, **args):
```
charn, line 206: I know the documentation says `**args`, but maybe for consistency we should use `**kwargs` here.

profiles/schema.py
```diff
@@ -370,6 +371,46 @@ def filter_by_nin_exact(self, queryset, name, value):
             return queryset.none()
 
 
+class AllowedDataFieldsMixin:
```
charn, line 374: Not sure if the correct place to subclass this would be on the graphene node classes or on the Django `Model` classes. :thinking:

open_city_profile/graphene.py
nicobav: This is true and I actually started about going for the model side. But then I thought since the restriction concerns the api, the restriction logic should be in the api not in the model 🤔 
But I'm open to this way of solution also if there are strong opinions towards this.

open_city_profile/graphene.py
nicobav: Very good proposion. Didn't notice it myself - obviously 😅

profiles/schema.py
nicobav: Justifying my decision; my thought was to lean more for the api side since I think model should know as little as possible for the api and its visibilities => handle the visibility checks/permissions in the api level not in model => schema/nodes

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

open_city_profile/graphene.py
charn: Don't know if it's good or not, but the solution which I suggested wouldn't need `_is_profile_query` method at all. It would apply anytime a model has `check_allowed_data_fields` set to `True` i.e. not just on those specific queries.

open_city_profile/graphene.py
```diff
@@ -178,3 +181,44 @@ def __init_subclass_with_meta__(
...
+                    "Allowed data field exception would occur: Service not identified. Field name: %s",
+                    field_name,
+                )
+
+            if not node.is_field_allowed_for_service(field_name, info.context.service):
+                if settings.ENABLE_ALLOWED_DATA_FIELDS_RESTRICTION:
+                    raise FieldNotAllowedError("Field is not allowed for service.")
+
+                logging.warning(
+                    "Allowed data field exception would occur. Field (%s) is not allowed for service %s. Field",
```
charn, line 219: There might be an extra `Field` at the end of this warning message.

nicobav: Nice catch 🎣

open_city_profile/graphene.py
nicobav: I changed those checks being in the model and thus be using the root in the middleware 👌

profiles/schema.py
nicobav: changed to model solution. It truly is more efficient that way 👌

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr494.api.dev.hel.ninja** :rocket::rocket::rocket:

open_city_profile/graphene.py
```diff
@@ -178,3 +182,35 @@ def __init_subclass_with_meta__(
...
+
+
+class AllowedDataFieldsMiddleware:
+
+    def resolve(self, next, root, info, **kwargs):
+
+        operation_type = info.operation.operation
+        if operation_type == OperationType.QUERY and getattr(
+            root, "check_allowed_data_fields", False
+        ):
```
charn, line 194: We probably should include all operation types (i.e. we shouldn't check if the operation is query). Checking only queries allows the service to access the fields through doing a mutation and querying every field though that.

```suggestion
        if getattr(root, "check_allowed_data_fields", False):
```

nicobav: Yes we should, again very good observation!

charn (approved):
Good work! LGTM! :+1: :shipit:

PR #499, opened 2024-05-02
fix: query suggestions are enabled when debugging

charn:
Also move DEBUG logic part of `ENABLE_GRAPHIQL` and `ENABLE_GRAPHQL_INTROSPECTION` into `settings.py`

Refs: HP-2351


voneiden (approved):
I like it

PR #505, opened 2024-07-04
fix: use getattr to access the original_error

nicobav:
GraphQlView's format_error is accessing to error's original_error. When the original_error is not present it caused the format_error itself to raise an error.
This add better handling for the original_error, using getattr, to avoid the above mentioned.

Refs HP-2485


terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr505.api.dev.hel.ninja** :rocket::rocket::rocket:

charn (approved):
LGTM! :+1: :shipit:

PR #507, opened 2024-07-18
HP-2490: Add login_methods field in ProfileNode

danipran:


profiles/keycloak_integration.py
```diff
@@ -8,7 +10,7 @@
 )
 from utils import keycloak
 
-_keycloak_admin_client = None
+_keycloak_admin_client: Optional[keycloak.KeycloakAdminClient] = None
```
charn, line 13: I guess it's a matter of taste, but since Python 3.10 you can also use:
```suggestion
_keycloak_admin_client: keycloak.KeycloakAdminClient | None = None
```

danipran: Yeah, no, that's definitely how it should be done with Python >=3.10.

profiles/schema.py
```diff
@@ -589,6 +598,25 @@ class Meta:
         "privileges to access this information.",
     )
 
+    def resolve_login_methods(self: Profile, info, **kwargs):
+        if info.context.user != self.user:
+            raise PermissionDenied(
+                "No permission to read login methods of another user."
+            )
+
+        amr = {info.context.user_auth.data.get("amr")}
```
charn, line 607: There's a small catch here. Tunnistamo returns the `amr`-claim as a string (which is not the standard way). Keycloak on the other hand will return this as a list. We could future-proof this so this doesn't break when we finally migrate over to the keycloak. e.g. [this is how it's done in linked events](https://github.com/City-of-Helsinki/linkedevents/blob/main/helevents/models.py#L36-L42).

Though once we migrate fully to keycloak we probably don't really need to check the amr-claim any more since there's no extra authentication methods available, which now exist on tunnistamo. :thinking:

charn: Maybe we can future-proof this and also add a note that amr-check can be removed once open-city-profile doesn't use tunnistamo anymore? :thinking:

danipran: I actually had it as a list first, but changed it based on how it was implemented in unit tests. Will add logic for handling both cases.

profiles/schema.py
```diff
@@ -575,6 +579,11 @@ class Meta:
         connection_class = ProfilesConnection
         filterset_class = ProfileFilter
 
+    login_methods = graphene.List(
+        graphene.String,
+        description="List of login methods that the profile has used to authenticate. "
+        "Only visible to the user themselves.",
+    )
```
danipran, line 586: Juha suggested using Graphene enums here instead of string. Will refactor this.

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr507.api.dev.hel.ninja** :rocket::rocket::rocket:

profiles/enums.py
```diff
@@ -41,3 +41,16 @@ class Labels:
         WORK = _("Work address")
         HOME = _("Home address")
         OTHER = _("Other address")
+
+
+class LoginMethodType(Enum):
+    PASSWORD = "password"
+    OTP = "otp"
+    SUOMI_FI = "suomi_fi"
+    HELSINKIAD = "helsinkiad"
```
charn, line 50: I think we can remove `HELSINKIAD` since at least by definition Helsinki AD users should not get a Helsinki profile.

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr507.api.dev.hel.ninja** :rocket::rocket::rocket:

profiles/schema.py
```diff
@@ -589,6 +601,26 @@ class Meta:
...
+
+        # For future software archeologists:
+        # This field was added to the API to support the front-end's need to know
+        # which login methods the user has used. It's only needed for profiles
+        # with helsinki-tunnus or Suomi.fi, so for other cases, save a couple
+        # API calls and return an empty list. There's no other reasoning for the
+        # logic here.
+        # Can remove this after Tunnistamo is no longer in use. Related ticket: HP-2495
+        if amr.intersection({"helsinki_tunnus", "heltunnistussuomifi", "suomi_fi"}):
+            return get_user_login_methods(self.user.uuid)
```
charn, line 620: Here (or in the function) we could check that the values which are returned are part of the enum. So only values which are part of the enum are returned and maybe add a `logger.warning(f"Loging methods found which are not part of the enum: {values_not_in_enum}")` for the values missing from the enum.

Tried this locally and since my account also had `recovery-authn-codes` I got the following error straight away:

```
 "Enum 'LoginMethodType' cannot represent value: 'recovery-authn-codes'",
```

PR #521, opened 2024-10-14
RATY-129 | Use DATABASE_PASSWORD env variable

nicobav:
This makes use of the DATABASE_PASSWORD env variable if it has been setupped.

Also bump commitlint to 9.18.0 to make it work with the .js config.

Refs RATY-129


tuomas777 (approved):
👍

PR #523, opened 2024-10-15
fix: actually raise exception

voneiden:
As pointed out by sonarcloud.

### DIFF
--- a/open_city_profile/settings.py
+++ b/open_city_profile/settings.py
@@ -301,36 +301,6 @@ GRAPHENE = {
     ],
 }
 
-if "SECRET_KEY" not in locals():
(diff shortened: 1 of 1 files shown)


nicobav:
I guess we want to do the same as in you did [here](https://github.com/City-of-Helsinki/linkedevents/pull/1002/files#diff-b05961a2c720fc36cf795ee2c2b036d9d5701c39783e0c6a4351210f255f8dd4L587-R587)?

voneiden:
Apparently I pushed my changes in a completely inappropriate place yesterday. Here it is now!

nicobav (approved):

PR #532, opened 2024-12-10
HP-2530 | Include more information for profile's login methods

charn:
This PR adds additional information fields related to a Profile's login methods (authentication methods in keycloak). Since `ProfileNode.loginMethods` is already in use, a new `ProfileNode.availableLoginMethods` field is added and the former is deprecated.

### DIFF
--- a/profiles/schema.py
+++ b/profiles/schema.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Iterable
 from itertools import chain
 
 import django.dispatch
@@ -57,7 +58,10 @@ from .connected_services import (
     download_connected_service_data,
 )
 from .enums import AddressType, EmailType, LoginMethodType, PhoneType
-from .keycloak_integration import delete_profile_from_keycloak, get_user_login_methods
+from .keycloak_integration import (
+    delete_profile_from_keycloak,
+    get_user_login_methods,
+)
 from .models import (
     Address,
     ClaimToken,
@@ -92,7 +96,9 @@ AllowedPhoneType = graphene.Enum.from_enum(
 AllowedAddressType = graphene.Enum.from_enum(
     AddressType, description=lambda e: e.label if e else ""
 )
-
+LoginMethodTypeEnum = graphene.Enum.from_enum(
+    LoginMethodType, description=lambda e: e.label if e else ""
+)
 
 """Provides the updated Profile instance as a keyword argument called `instance`."""
 profile_updated = django.dispatch.Signal()
@@ -438,6 +444,16 @@ class AddressNode(ContactNode):
             )
 
 
+class LoginMethodNode(graphene.ObjectType):
+    method = LoginMethodTypeEnum(required=True, description="The login method used.")
+    created_at = graphene.DateTime(
+        description="Time when the login method was created or edited."
+    )
+    user_label = graphene.String(
+        description="User-friendly label for the login method."
+    )
+
+
 class VerifiedPersonalInformationAddressNode(graphene.ObjectType):
     street_address = graphene.String(
         required=True, description="Street address with possible house number etc."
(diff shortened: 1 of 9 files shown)


terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr532.api.dev.hel.ninja** :rocket::rocket::rocket:

profiles/schema.py
```diff
@@ -605,6 +625,42 @@ class Meta:
...
+        which login methods the user has used. It's only needed for profiles
+        with helsinki-tunnus or Suomi.fi, so for other cases, save a couple
+        API calls and return an empty list. There's no other reasoning for the
+        logic here.
+        Can remove this after Tunnistamo is no longer in use. Related ticket: HP-2495
+        """
+        return amr.intersection({"helsinki_tunnus", "heltunnistussuomifi", "suomi_fi"})
+
+    @staticmethod
+    def _get_loging_methods(user_uuid, extended=False) -> Iterable:
```
danipran, line 642:
```suggestion
    def _get_login_methods(user_uuid, extended=False) -> Iterable:
```

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr532.api.dev.hel.ninja** :rocket::rocket::rocket:

profiles/tests/test_gql_my_profile_query.py
```diff
@@ -634,25 +646,42 @@ def mock_return(*_, **__):
...
+                }
             }
         }
     """
     executed = user_gql_client.execute(
         query, auth_token_payload={"amr": amr_claim_value}, service=service
     )
     assert "errors" not in executed
-    assert set(executed["data"]["myProfile"]["loginMethods"]) == {
+    assert executed["data"]["myProfile"]["loginMethods"] == [
```
danipran, line 661: Can the order change here or is it always the same? And if it can change, does the order matter? The reason I originally used `set` here instead of a `list` was to ignore the order of the items and just compare the values (because `["foo", "bar"] != ["bar", "foo"]`)

profiles/schema.py
```diff
@@ -605,6 +625,42 @@ class Meta:
...
+        ):
+            logger.warning(
+                "Found login methods which are not part of the LoginMethodType enum: %s",  # noqa: E501
+                unknown_login_methods,
+            )
+
+        if extended:
+            return login_methods_in_enum
+        else:
+            return {val["method"] for val in login_methods_in_enum}
```
danipran, line 662:
```suggestion
            return [val["method"] for val in login_methods_in_enum]
```
...I think?

profiles/schema.py
danipran, line 642: Avoid boolean trap:
```suggestion
    def _get_loging_methods(user_uuid, *, extended=False) -> Iterable:
```

danipran (approved):
LGTM, left a couple suggestions :+1:

profiles/tests/test_gql_my_profile_query.py
charn: The order should be stable at least in the tests since database doesn't affect the order. Order doesn't matter, so in that sense it could be a `set`. Though a `set` could potentially hide a value existing multiple times in the data.

terovirtanen:
**HELSINKI-PROFILE-API branch is deployed to platta: https://helsinki-profile-pr532.api.dev.hel.ninja** :rocket::rocket::rocket:

profiles/tests/test_gql_my_profile_query.py
danipran: Yeah, the potential duplicates were also a reason why I used sets quite a lot, as you probably noticed. My reasoning back then was that the duplicates wouldn't probably really matter, so I used sets for deduplication.

PR #561, opened 2025-08-26
HP-2926 | Update Python to 3.12

charn:
Bump Python and `pre-commit` related tools. Dockerfile is simplified a bit and unified with other projects. Local developer experience is also maintained by using the old `appuser` for local development (with uid 1000). Rename docker compose related files according to the current recommendations.


voneiden (approved):
:shipit:

